PR #4, opened 2019-02-26
Add Python 3+ type-annotations

y2k-shubham:
Use [`typing` module](https://docs.python.org/3/library/typing.html) to add Type-Annotations as per Python 3+

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ pip-selfcheck.json
 
 # vscode
 .vscode
+
+# IntelliJ
+.idea/*
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -1,93 +1,101 @@
+from datetime import datetime
+from typing import Dict, Any, List, Union, Callable
+
 from airflow import DAG, configuration
+from airflow.models import BaseOperator
 from airflow.utils.module_loading import import_string
 
 from dagfactory import utils
 
 # these are params only used in the DAG factory, not in the tasks
-SYSTEM_PARAMS = ["operator", "dependencies"]
+SYSTEM_PARAMS: List[str] = ["operator", "dependencies"]
 
 
 class DagBuilder(object):
-    def __init__(self, dag_name, dag_config, default_config):
-        self.dag_name = dag_name
-        self.dag_config = dag_config
-        self.default_config = default_config
 
-    def get_dag_params(self):
+    def __init__(self,
+                 dag_name: str,
+                 dag_config: Dict[str, Any],
+                 default_config: Dict[str, Any]) -> None:
+        self.dag_name: str = dag_name
+        self.dag_config: Dict[str, Any] = dag_config
+        self.default_config: Dict[str, Any] = default_config
+
+    def get_dag_params(self) -> Dict[str, Any]:
         """
         Merges default config with dag config, sets dag_id, and extropolates dag_start_date
 
         :returns: dict of dag parameters
         """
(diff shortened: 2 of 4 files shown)


y2k-shubham:
**@ajbosco** thanks a ton! (that was quick)
Heads up for this excellent project; it elegantly throws light on the _right way of constructing DAGs_.
.
At **@Zomato** we aren't using `dag-factory` at the moment but we are actively exploring it and other similar solutions (like [etsy/boundary-layer](https://github.com/etsy/boundary-layer)) for inspiration.

Jaikant:
@y2k-shubham Which one did you choose at Zomato - dag-factory or boundary-layer?

y2k-shubham:
We didnt picky [`etsy/boundary-layer`](https://github.com/etsy/boundary-layer) because
1. adoption seemed risky since it hadn't caught community traction (the last thing one wants is to end up running a platform built on an abandoned library / framework)
2. merely scratching it on the surface revealed several shortcomings ([#31](https://github.com/etsy/boundary-layer/issues/31), [#33](https://github.com/etsy/boundary-layer/issues/33), [#43](https://github.com/etsy/boundary-layer/issues/43)) that would've required us to customize it (since community support was non-existent)
.
and as far as `dag-factory` is concerned, it is more of an idea (for a general approach) than a full-fledged solution
.
so taking inspiration from `dag-factory` and also analyzing our requirements, we ended up implementing our custom solution (took us ~ 1 year to evolve and mature the platform) where
1. we have JSON configs to generate DAGs (1 JSON per DAG)
2. generated DAGs can't be arbitrary, rather they have a well defined structure (in terms on operators, and their connections)
3. generated DAGs are one of several possible (small no of) types or categories; each having a specific structure in terms of operator and their connections
4. because of point 3. above, we now have JSONs with pre-defined structure (each category of DAGs has a specific structure of JSON) that can be validated beforehand. Also since structure of DAGs is known ahead of time (needs not be inferred from JSON), we have a `DagBuilder` class per category of DAG that generates DAGs of that category while passing appropriate args (read from JSON config) to operators
5. Much like `dag-factory` (and unlike `boundary-layer`), we dont perform any code-generation. Rather DAGs are materialized in memory only
.
why we picked this approach
1. we had to support a large no of (150+) internal users who had little or no knowledge of underlying Airflow platform (over 90% of users aren't even developers, they are very good business analysts who know SQL like the back of their hands). So asking them to specify operators for their DAGs was dumb (and error prone). Plus by not putting many Airflow-specific info in JSON configs (and abstracting out Airflow details from end users), we have the liberty to move to a different platform in future with minimal effort
2. We wanted to have tight control over the logic of DAGs themselves partly because we were the owners of the platform and also because our users weren't well versed with Airflow. So by writing the DAG generation code ourselves, we made sure that their were no surprises (other than those where JSON configs themselves were broken)
3. Since JSONs had well defined structure. we had the option to run validations on them (so that our DAG generation code doesn't error out). Validations could be like all necessary keys are present in JSON, all values of expected types etc.
.
One thing to point out, by employing this approach, we were also able to achieve dependednt DAGs (one DAG triggers another set of DAGs upon completion) simply by having user specify in their JSONs `"depends_on": ["dag_a", "dag_b"]` (now this DAG will be triggered by either of `dag_a` or `dag_b` (whichever finishes later)

Jaikant:
Awesome. Sounds like you did all the right things. I am still in two minds about adopting airflow and somewhere I feel dagster-io has better abstractions built in.

PR #45, opened 2020-05-11
Permit setting 'None' as valid schedule interval

SimonArnu:
When defining DAGs the schedule_interval param is read in as string.
This way, setting 'None' as schedule_interval is not possible, as it is no valid cron expression.
With this patch it is possible to define DAGs that never get scheduled automatically.


ajbosco:
I think if you set it to `schedule_interval: null` then PyYaml will interpret it as `None` correctly.

SimonArnu:
Ok, setting `schedule_interval: null` works as well as `schedule_interval:`.
Then this pull request is obsolete.

PR #46, opened 2020-05-24
Added execution_timeout parse at task level and tags parameter at dag level.

jelther:
It wasn't possible to set the execution timeout at task level since timedelta objects were not parsed when an operator was instantiated.
Thus, added a new parameter called "execution_timeout_secs" that adds this parameter to the operators.

I've noticed that the "sla" parameter on the BaseOperator is not being parsed too. Will change in a future commit.


jelther:
Added also the "tags" parameter to be able to use tags on newer versions of airflow.

ajbosco:
Thanks!

PR #48, opened 2020-06-11
Fix 'tags' error

withsmilo:
The default value of 'tags' should be 'None'. See https://github.com/apache/airflow/blob/c41192fa1fc5c2b3e7b8414c59f656ab67bbef28/airflow/models/dag.py#L236.


ajbosco:
Thanks!

PR #52, opened 2020-08-07
Added support for KubernetesPodOperator

jelther:
Also bumped airflow version to 1.10.11.


dagfactory/dagbuilder.py
```diff
@@ -2,11 +2,25 @@
 from datetime import timedelta, datetime
 from typing import Any, Callable, Dict, List, Union
 
+import os
+
 from airflow import DAG, configuration
+from airflow.models import Variable
+
+# kubernetes operator
+from airflow.kubernetes.secret import Secret
```
ajbosco, line 11: I think these imports should be `airflow.contrib.kubernetes` to keep it backwards compatible with earlier version of Airflow.

tests/test_dagfactory.py
```diff
@@ -61,6 +67,7 @@ def test_load_config_valid():
             },
         },
         "example_dag2": {
+            "doc_md_file_path" : "./fixtures/mydocfile.md",
```
ajbosco, line 70: A few tests are failing with this error:

```
FileNotFoundError: [Errno 2] No such file or directory: './fixtures/mydocfile.md'
```

Do these need to be fully qualified? `os.path.join(here, "fixtures/mydocfile.md")`

Issue #42, opened 2020-04-14
start_date and end_date get their time component removed


swapnil-pandey:
While using the dag-factory to create dags on Apache airflow, I noticed a very strange behavior. Suppose while creating a dag, input start_date is "2020-04-14T16:20:30.000Z". On airflow UI, it is shown as "2020-04-14T00:00:00.000Z" (Notice how the time component is removed).

I did a little bit of searching and I found out this issue to be due to this piece of code in `utils.py`:
``` 
 if isinstance(date_value, date):
    return datetime.combine(date=date_value, time=datetime.min.time()).replace(
        tzinfo=local_tz
    ) 
if isinstance(date_value, datetime):
    return date_value.replace(tzinfo=local_tz)
```

If I pass a datetime.datetime object as date_value, then it will match with the first if and that will remove the time component of my input. 
Can you suggest a workaround if I am doing something wrong?

ajbosco:
Looks like a bug, I'll try to get to it this weekend, but feel free to submit a PR if it is blocking you.

PR #53, opened 2020-08-20
Check date_value types explicitly in get_datetime

coopergillan:
__What__

Check the type of `date_value` passed to `get_datetime` in order to preserve the time component.

__Why__

Since `datetime.datetime` is a subclass of `datetime.date, using `isinstance on and instance of either class will return True. Using `type` will return the exact class.

Closes: #42


coopergillan:
Close #54 without merging if this one is preferred.

coopergillan:
Closing in favor of #54.

coopergillan:
The more I thought about it, I like the other approach more.

PR #54, opened 2020-08-20
Check for datetime first in get_datetime util

coopergillan:
__What__

Check whether `date_value` is `datetime.datetime` object before checking for a `datetime.date` object.

__Why__

Since `datetime.datetime` objects are subclasses of date objects, the first check will wrongly assume that a `datetime.datetime` parameter is a `datetime.date` class.

__A Note__

#53 was the original proposal, but required overriding the `nidiomatic-typecheck` pylint check. Either one is fine, but that one is definitely more explicit.

Closes: #42


coopergillan:
Close #53 without merging if this one is preferred.

coopergillan:
Closing in favor of #53.

coopergillan:
The more I thought about it, the more I liked this approach.

PR #71, opened 2021-01-07
Handle "None" more flexibly as the value of schedule_interval

withsmilo:
If the yaml file we created looks like this:
```
test1: None
test2: null
test3: "None"
```
YAML Loader reads as follows.
```
{'test1':'None','test2': None,'test3':'None'}
```
In this case, it would be better to handle "None" more flexibly
because Airflow doesn't allow "None" as schedule_interval.
Please refer https://airflow.apache.org/docs/apache-airflow/1.10.10/scheduler.html#dag-runs.

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -67,6 +67,12 @@ class DagBuilder:
             raise "Failed to merge config with default config" from err
         dag_params["dag_id"]: str = self.dag_name
 
+        if (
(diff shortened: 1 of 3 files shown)


ajbosco (approved):
Thanks!

withsmilo:
@ajbosco 
Thanks for quick merging! Would you please release a patch version?

ajbosco:
> @ajbosco
> Thanks for quick merging! Would you please release a patch version?

@withsmilo just cut a new release

PR #72, opened 2021-01-11
Yujia/add customize operator sample

yogyang:
Hey!

The lib is great and I was testing some use cases to use it, I changed some config to run docker with airflow 2-0-0  and add some examples for customized operators.

Cheers


ajbosco (approved):
Thanks!

PR #73, opened 2021-01-19
Airflow sets different default value of the 'description' depending on the version

withsmilo:
* Airflow 1.10.10 uses `''` as default value. https://github.com/apache/airflow/blob/1.10.10/airflow/models/dag.py#L218
* Airflow 1.10.11 uses `None` as default value. https://github.com/apache/airflow/blob/1.10.11/airflow/models/dag.py#L221


ajbosco:
Looks like it is failing a few tests.

withsmilo:
Sorry about that. I'm trying to fix my error.

withsmilo:
@ajbosco 
passed! Would you please review this one? :)

ajbosco (approved):
Thanks!

PR #81, opened 2021-05-10
Task groups and dict dag config

burakyilmaz321:
Hi @ajbosco ✋ 

This PR introduces python dicts as configuration file, and task groups.

* Adding an option for initializing `DagFactory` instances with python dictionaries, instead of YAML file path, allows users to read, edit, merge configurations on the fly, and create a dag with it.
* TaskGroups is a great feature coming with Airflow 2. Added new special keywords `task_groups` to dag, and `task_group_name` to task sections.

It would be great if you can review and tell me what you think 😄


ajbosco:
@burakyilmaz321 Looks great, but it isn't passing the formatter. It would also be awesome if you could add a test for the new method.

burakyilmaz321:
@ajbosco sorry for delay. All unit tests and formatters passes now. I also added a couple of unit tests for various cases. Please let me know if you want me to do anything else.

ajbosco (approved):
Awesome. Thanks!

bodhwani:
@burakyilmaz321 Can you show an example of how task groups can be used in yaml file?

avico78:
> @burakyilmaz321 Can you show an example of how task groups can be used in yaml file?

An example would be really helpfull, thanks!

PR #90, opened 2021-07-19
Allow callbacks from python modules

psg2:
Hi, 

I would like to contribute this change to allow users to specify a callback function as a module the same way we have for the operator keyword.


ajbosco:
@psg2 looks like CI is failing. Can you address the errors when you get a chance?

psg2:
> @psg2 looks like CI is failing. Can you address the errors when you get a chance?

Sure. One of the errors right now is that the function has too many branches because I basically added a bunch of if conditions. Do you have any suggestions on how to simplify all of those conditionals?

I thought about something like this, what do you think?

```py
        def mutate_dagrun_timeout_sec(params: Dict[str, Any]):
            params["dagrun_timeout"]: timedelta = timedelta(
                seconds=dag_params["dagrun_timeout_sec"]
            )
            del params["dagrun_timeout_sec"]

        mutations = [
            {
                "handler": mutate_dagrun_timeout_sec,
                "required_keys": ["default_args.dagrun_timeout_sec"],
            }
        ]

        for mutation in mutations:
            for required_key in mutation.get("required_keys"):
                if utils.check_dict_key(dag_params, required_key):
                    mutation["handler"](dag_params)
```

ajbosco:
> > @psg2 looks like CI is failing. Can you address the errors when you get a chance?
> 
> Sure. One of the errors right now is that the function has too many branches because I basically added a bunch of if conditions. Do you have any suggestions on how to simplify all of those conditionals?
> 
> I thought about something like this, what do you think?

That looks good to me, but if it becomes too difficult I'm definitely ok with just disabling that pylint rule.

psg2:
> > > @psg2 looks like CI is failing. Can you address the errors when you get a chance?
> > 
> > 
> > Sure. One of the errors right now is that the function has too many branches because I basically added a bunch of if conditions. Do you have any suggestions on how to simplify all of those conditionals?
> > I thought about something like this, what do you think?
> 
> That looks good to me, but if it becomes too difficult I'm definitely ok with just disabling that pylint rule.

Fixed, but this new approach didn't work so well too. 😢 

After refactoring it hit another lint error `dagfactory/dagbuilder.py:213:4: R0914: Too many local variables (16/15) (too-many-locals)` and decided to just remove the lint rule on the function. If you prefer, I can rollback to the previous version with a lot of ifs and remove the rule as well.

ajbosco (approved):

ajbosco:
Thank you!

PR #116, opened 2021-12-10
Add environment variables support for python callable file

indyyyyy:
**Problem:**
It seems that you cannot use environment variables or Airflow variables when you set dag params like `python_callable_file` or `on_success_callback_file`. With env var in the file path, the get_python_callable function will consider this path as non absolute path or it will return a `No such file or directory` error.

**Test:**
Local DAG Loader Test (e.g. python `your-dag-file.py` which process my test-dag.yaml)

**Goals:**
Add support for callable file path and allow developpers to use environment variables and not get error on `get_python_callable` function. It just translate `python_callable_file`'s environment variable so if there is.

### DIFF
--- a/dagfactory/utils.py
+++ b/dagfactory/utils.py
@@ -118,6 +118,8 @@ def get_python_callable(python_callable_name, python_callable_file):
     :type: callable
     """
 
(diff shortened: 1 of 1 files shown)


ajbosco:
Thanks!

Issue #104, opened 2021-09-06
Deserialise arbitrary types in python


dinigo:
Most part of params that the DAG and the Operators and Sensors underneath take are basic/raw types (string, int, bool, float) and basic objects (lists, dicts). But from time to time we require some python object. For example:
* ExternalTaskSensor [execution_delta](https://github.com/apache/airflow/blob/main/airflow/sensors/external_task.py#L67) is datetime.timedelta type.
* PythonOperator [python_callable](https://github.com/apache/airflow/blob/main/airflow/operators/python.py#L88) is of Callable type

An example for the `ExternalTaskSensor`

```yaml
example_external_task_sensor:
  default_args:
    start_date: 2021-09-01
  schedule_interval: @daily
  tasks:
    external_task_sensor:
      operator: airflow.sensors.external_task.ExternalTaskSensor
      external_dag_id: example_upstream_dependency_dag
      external_task_id: example_upstream_dependency_task
      execution_delta: !!python/object/apply:datetime.timedelta [0, 300]
```
And also an example for deserialising a reference to the callable straight from the yaml, lets suppose we are instantiating this yaml from a file  `main_dag_file.py` that also contains a python function we want to call (so that it belongs to the PYTHONPATH):

```python
# main_dag_file.py
def add_two_numbers(first_operator: int, second_operator: int) -> int:
  return first_operator + second_operator
```

```yaml
example_call_python_code:
  default_args:
    start_date: 2021-09-01
  schedule_interval: @daily
  tasks:
    example_python_task:
      operator: airflow.operators.python.PythonOperator
      python_callable: !!python/name:main_dag_file.add_two_numbers
      op_wkargs:
        first_operator: 3
        second_operator: 2
```

In addition sometimes it's useful to both:
* Have your DAG as a config file with it's jinja templating ready for airflow.
* Have a render engine on top to be able to make references, modify strings, pass function return as a param for an operator.

```python
# main_dag_file.py
def decide_executor() -> str:
  environment = os.getenv('EXECUTION_ENVIRONMENT')
  if environment == 'prod':
    executor = 'kubernetes.KubernetesPodOperator'
  else:
    executor = 'bash.BashOperator'
  return 'airflow.operators.' + executor
```
```
# pass environment variables to the function
example_call_python_code:
  default_args:
    start_date: 2021-09-01
  schedule_interval: @daily
  tasks:
    example_python_task:
      operator: !!python/object/main_dag_file.decide_executor()
[...]

PR #123, opened 2022-02-02
Allow providing python callback through config

dinigo:
There's a couple cases we might want to provide a python callback through config.
First one when we build the config ourselves dinamically and we provide a dict to the
Dagfactory. Another is loading straight from pure yaml like so
```yaml
my_dag:
  tasks:
    run_my_function:
      operator: airflow.operators.python.PythonOperator
      python_callable: !!python/name:my_module.my_function
```

This is linked to the issue #104


ajbosco (approved):

dinigo:
@ajbosco, thanks for allowing people contribute to your project. And thanks for putting the time in maintaining it :)

PR #124, opened 2022-02-07
fix: add params to DAG initialization

aramakrishnan19:
Allow params to be specified in the yaml

YAML definition.

```yaml
params:
    key: value
```
This translates to 
```python
DAG(...,params={...})
```
Reference : https://github.com/apache/airflow/blob/main/airflow/models/dag.py#L229


ajbosco (approved):

ajbosco:
@aramakrishnan19 Thanks, I'll cut a release later tonight!

PR #129, opened 2022-04-28
Add support for python_callable_file for PythonSensor

arnekaiser-otto:


dagfactory/dagbuilder.py
```diff
@@ -194,14 +195,14 @@ def make_task(operator: str, task_params: Dict[str, Any]) -> BaseOperator:
...
             raise Exception(f"Failed to import operator: {operator}") from err
         try:
-            if operator_obj in [PythonOperator, BranchPythonOperator]:
+            if operator_obj in [PythonOperator, BranchPythonOperator, PythonSensor]:
                 if (
                     not task_params.get("python_callable")
                     and not task_params.get("python_callable_name")
                     and not task_params.get("python_callable_file")
                 ):
                     raise Exception(
```
ajbosco, line 204:
```suggestion
                    # pylint: disable=line-too-long
                    raise Exception(
```

ajbosco (approved):

dagfactory/dagbuilder.py
```diff
@@ -10,6 +10,7 @@
 from airflow.contrib.operators.kubernetes_pod_operator import KubernetesPodOperator
 from airflow.models import BaseOperator
 from airflow.operators.python_operator import PythonOperator, BranchPythonOperator
+from airflow.sensors.python import PythonSensor
```
ajbosco, line 13: Can you move this to line 42? `PythonSensor` wasn't introduced until later and I'd like to maintain backwards compatibility with older Airflow versions if possible

arnekaiser-otto: Yes sure, sorry for number of tries...

arnekaiser-otto: Another question with a similar background, I face the callable Problem with the google pubsub sensor as well, the sensor has a python class for minimal processing of the messages as default or you can call an own function, both does not work out of the box, do you have an idea how to solve that?

ajbosco, line 13:
```suggestion
```

PR #144, opened 2022-10-21
Add function to scan recursively searching for yaml DAGs

dinigo:
Hello! I've got this recurring use case where I scan my  DAGs folder recursively looking for DAGs defined in yaml files.

B default this prefixes are `.yml` and `.yaml`. But I found that filtering out files not containing DAGs is convenient. You can do so setting the. `prefix=['dag.yaml']` to match files like `my_dinigo_dag.yaml` and `my_ajbosco.dag.yaml`. But not `operator_config.yaml`.  The scan path defaults to the `dag folder` extracted from running airflow configuration.

I am using this on my package, but it is a nice addition to the library and I thought it belonged here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,15 @@ dag_factory.generate_dags(globals())
 
 And this DAG will be generated and ready to run in Airflow!
 
+If you have several configuration files you can improt them like this:
+
+```python
+# 'airflow' word is required for the dagbag to parse this file
+from dagfactory import load_yaml_dags
+
+load_yaml_dags(globals_dict=globals(), suffix=['dag.yaml'])
+```
+
 ![screenshot](/img/example_dag.png)
 
 ## Notes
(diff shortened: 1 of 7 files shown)


dinigo:
@ajbosco there are several preexisting  linting errors that precent the pipeline from succeeding. Should I fix them even thoug they don't come from this PR?

README.md
```diff
@@ -67,6 +67,15 @@ dag_factory.generate_dags(globals())
 
 And this DAG will be generated and ready to run in Airflow!
 
+If you have several configuration files you can improt them like this:
```
ajbosco, line 70:
```suggestion
If you have several configuration files you can import them like this:
```

ajbosco:
@dinigo I think I've got the other errors fixed now so you should be able to update your branch.

dinigo:
Moved to the `dagfactory` module to avoid circular import

ajbosco (approved):

vanducng:
Looks great, can you release a new tag @ajbosco

ajbosco:
> Looks great, can you release a new tag @ajbosco

@vanducng just cut a new release.

PR #157, opened 2023-03-21
Allow inheritance of KubernetesPodOperator

### LINKED ISSUE #156: dag-factory does not support operator inheriting from KubernetesPodOperator
I have a custom operator which inherits from the `KubernetesPodOperator` and sets some defaults, mainly for ease of use. It works well in simple situations where I don't need any Kubernetes constructs, but I now need to mount a volume to the pod and I'm having trouble.

The error I'm getting is this:
```
airflow.exceptions.AirflowException: Expected <class 'kubernetes.client.models.v1_volume_mount.V1VolumeMount'>, got <class 'str'>
```

Which means the `DagBuilder` is not handling the `volume_mounts` entry in the config file. I believe this is happening because it checks whether the object is a `KubernetesPodOperator` [here](https://github.com/ajbosco/dag-factory/blob/bdb2b773732f02bb55a5dd96fad807b7ad26d691/dagfactory/dagbuilder.py#L375), but this returns false if the operator is a child of it.

I'm going to try to open a PR to make the condition return true also if the operator is a child of the `KubernetesPodOperator`.

dimon222:
This fixes #156 
TBH potentially the same condition should be used for all operators to allow custom implementations of sensors, etc

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -372,7 +372,7 @@ class DagBuilder:
                     del task_params["response_check_lambda"]
 
             # KubernetesPodOperator
-            if operator_obj == KubernetesPodOperator:
+            if isinstance(operator_obj, KubernetesPodOperator):
                 task_params["secrets"] = (
                     [Secret(**v) for v in task_params.get("secrets")]
                     if task_params.get("secrets") is not None


ajbosco:
Awesome. I'm traveling this week but will get this merged and released when I'm back at a computer. Thanks!

ajbosco (approved):

PR #159, opened 2023-03-29
Allow inherited operators

### LINKED ISSUE #156: dag-factory does not support operator inheriting from KubernetesPodOperator

dimon222:
This is a revised approach that didn't work in #157  to resolve #156
Since import_string imports a class, not an instance of class, it caused problem that broke the original mechanic. This PR is relying on another less commonly used method `issubclass` that supposed to offer the behavior that is actually expected.

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -35,6 +35,13 @@ try:
 except ImportError:
     from airflow.providers.common.sql.sensors.sql import SqlSensor
 
+# python sensor was moved in Airflow 2.0.0
+try:
+    from airflow.sensors.python import PythonSensor
+except ImportError:
+    from airflow.contrib.sensors.python_sensor import PythonSensor
+
+
 # k8s libraries are moved in v5.0.0
 try:
     from airflow.providers.cncf.kubernetes import get_provider_info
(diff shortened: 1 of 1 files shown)


dimon222:
Hm, I don't understand why 3.8 tests error out with error 255 since the test suite did pass.
And for 3.7 - exception is unclear, since the logic behaves exactly what exception says. Perhaps something about python 3.7.x tree specifically...

dimon222:
Fixed handling of Airflow 1.10.x which apparently had broken `PythonSensor` even before my PR, that caused the new logic to not function properly and complain about `None` object appear among tuple for comparison.

ajbosco (approved):
Nice I'll cut a release for this change tomorrow.

PR #163, opened 2023-04-10
Added render_template_as_native_obj support

matveykortsev:
Added support and tests for render_template_as_native_obj.
It helps to use Jinja Templates in classic operators same as in dynamic ones, when you need python objects instead of strings


matveykortsev:
> 

Oh rly...

ajbosco (approved):

Issue #151, opened 2022-12-08
new feature after airflow 2.4.0 - can dag-factory support Datasets?


JiaruiShenOZ:
[Datasets and data-aware scheduling in Airflo](https://docs.astronomer.io/learn/airflow-datasets)

Schedule uses Datasets not time table.

pankajastro:
PR: https://github.com/astronomer/dag-factory/pull/171 has been merged and released long back so closing it.

PR #171, opened 2023-07-17
Support for Airflow Datasets

cicerojmm:
Implementation of Airflow Datasets (data-aware scheduling) functionality that became available from version 2.4.0.

In this implementation it is possible to build Datasets in two ways:
- The first is directly referencing the URI of the dataset in the appropriate parameters of both the producing tasks (outlets) and the consuming DAGs (schedule)
- the second is to have a specific YAML file to declare datasets (URI and Names) and thus reference this file indicating only the names of the datasets to be used by the producing tasks and consuming DAGs, as the URI will be selected automatically.

Referencied Issue: #151

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -112,6 +112,12 @@ else:
     XComArg = None
 # pylint: disable=ungrouped-imports,invalid-name
 
+if version.parse(AIRFLOW_VERSION) >= version.parse("2.4.0"):
+    from airflow.datasets import Dataset
+else:
+    Dataset = None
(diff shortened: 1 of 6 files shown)


ericbuzato:
This feature is going to be really cool, I can't wait to use it in my pipelines! ☝️

ciceromoura:
@ajbosco 
I added tests to increase coverage, I believe now the merge will pass

ajbosco:
Thanks!

karguru-aws:
Does the code factory support AWS Operators like Glue ?

PR #174, opened 2023-09-20
Cast callbacks to functions when set with default_args on task groups

Baraldo:
Hello!

When used in the task group's default_args, callbacks are passed as strings to the tasks, which ends up generating failures.

The idea is to perform the same parse used in `make_task` (it is necessary to check whether the default_args attribute is defined, since it is optional and defaults to None) and perhaps in the future modularize this conversion functionality in order to cover the widespread use of default_args.


pankajastro:
Hey @Baraldo, Thank you for the PR. Could you please rebase and fix the merge conflict?

pankajastro:
closing this favour of https://github.com/astronomer/dag-factory/pull/235

PR #184, opened 2024-03-12
Build DagFactory DAGs when there is an invalid YAML in the DAGs folder

quydx:
Before, when the Airflow DAGs folder had an invalid YAML file, no DagFactory DAGs would be loaded. This PR changes this behaviour to log any invalid YAML file paths but render valid DagFactory YAML-based DAGs.

Co-authored-by: Tatiana Al-Chueyr <tatiana.alchueyr@gmail.com>

### DIFF
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -178,9 +178,7 @@ def load_yaml_dags(
         suffix = [".yaml", ".yml"]
     candidate_dag_files = []
     for suf in suffix:
-        candidate_dag_files = list(
-            chain(candidate_dag_files, Path(dags_folder).rglob(f"*{suf}"))
-        )
+        candidate_dag_files = list(chain(candidate_dag_files, Path(dags_folder).rglob(f"*{suf}")))
     for config_file_path in candidate_dag_files:
         config_file_abs_path = str(config_file_path.absolute())
         logging.info("Loading %s", config_file_abs_path)
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -60,10 +60,13 @@ class DagFactory:
 
             yaml.add_constructor("!join", __join, yaml.FullLoader)
 
-            config: Dict[str, Any] = yaml.load(
-                stream=open(config_filepath, "r", encoding="utf-8"),
-                Loader=yaml.FullLoader,
-            )
+            with open(config_filepath, "r", encoding="utf-8") as fp:
+                yaml.add_constructor("!join", __join, yaml.FullLoader)
+                config_with_env = os.path.expandvars(fp.read())
+                config: Dict[str, Any] = yaml.load(
+                    stream=config_with_env,
+                    Loader=yaml.FullLoader,
+                )
         except Exception as err:
             raise DagFactoryConfigException("Invalid DAG Factory config file") from err
         return config
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -178,9 +178,15 @@ def load_yaml_dags(
         suffix = [".yaml", ".yml"]
     candidate_dag_files = []
     for suf in suffix:
-        candidate_dag_files = chain(candidate_dag_files, Path(dags_folder).rglob(f"*{suf}"))
-
+        candidate_dag_files = list(chain(
+            candidate_dag_files, Path(dags_folder).rglob(f"*{suf}")
+        ))
+    logging.info(candidate_dag_files)
     for config_file_path in candidate_dag_files:
         config_file_abs_path = str(config_file_path.absolute())
-        DagFactory(config_file_abs_path).generate_dags(globals_dict)
-        logging.info("DAG loaded: %s", config_file_path)
+        logging.info(f"Loading {config_file_abs_path}")
+        try:
+            DagFactory(config_file_abs_path).generate_dags(globals_dict)
+            logging.info("DAG loaded: %s", config_file_path)
+        except Exception as e:
+            logging.warning(f"Failed to load dag from {config_file_path}: {e}")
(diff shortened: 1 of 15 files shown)


anovis (approved):

dagfactory/dagfactory.py
```diff
@@ -190,11 +190,14 @@ def load_yaml_dags(
...
+        candidate_dag_files = list(chain(
             candidate_dag_files, Path(dags_folder).rglob(f"*{suf}")
-        )
-
+        ))
     for config_file_path in candidate_dag_files:
         config_file_abs_path = str(config_file_path.absolute())
-        DagFactory(config_file_abs_path).generate_dags(globals_dict)
-        logging.info("DAG loaded: %s", config_file_path)
+        logging.info(f"Loading {config_file_abs_path}")
```
kaxil, line 198:
```suggestion
        logging.info("Loading %s, config_file_abs_path")
```

kaxil (approved):

tatiana:
@quy196hp Thank you very much for your contribution, and I am sorry for the delay in giving feedback.

Please, could you address the linting issue, so the checks pass?
```
==> Formatting with Black
would reformat /home/runner/work/dag-factory/dag-factory/dagfactory/dagfactory.py

Oh no! 💥 💔 💥
1 file would be reformatted, 5 files would be left unchanged.
make: *** [Makefile:38: fmt-check] Error 1
```

dagfactory/dagfactory.py
```diff
@@ -195,7 +195,7 @@ def load_yaml_dags(
         ))
     for config_file_path in candidate_dag_files:
         config_file_abs_path = str(config_file_path.absolute())
-        logging.info(f"Loading {config_file_abs_path}")
+        logging.info("Loading %s, config_file_abs_path")
```
DavidTurland, line 198: I think this should be:
```python
logging.info("Loading %s", config_file_abs_path)
```
As is the log line will be(approx)
`INFO:root:Loading %s, config_file_abs_path`

dagfactory/dagfactory.py
```diff
@@ -190,11 +190,14 @@ def load_yaml_dags(
...
+        candidate_dag_files = list(chain(
             candidate_dag_files, Path(dags_folder).rglob(f"*{suf}")
-        )
-
+        ))
     for config_file_path in candidate_dag_files:
         config_file_abs_path = str(config_file_path.absolute())
-        DagFactory(config_file_abs_path).generate_dags(globals_dict)
-        logging.info("DAG loaded: %s", config_file_path)
+        logging.info("Loading %s, config_file_abs_path")
```
kaxil, line 198:
```suggestion
        logging.info("Loading %s", config_file_abs_path)
```

dagfactory/dagfactory.py
kaxil: You are right, good catch

dagfactory/dagfactory.py
```diff
@@ -190,11 +190,14 @@ def load_yaml_dags(
...
     for config_file_path in candidate_dag_files:
         config_file_abs_path = str(config_file_path.absolute())
-        DagFactory(config_file_abs_path).generate_dags(globals_dict)
-        logging.info("DAG loaded: %s", config_file_path)
+        logging.info("Loading %s", config_file_abs_path)
+        try:
+            DagFactory(config_file_abs_path).generate_dags(globals_dict)
+            logging.info("DAG loaded: %s", config_file_path)
+        except Exception as e:
+            logging.warning(f"Failed to load dag from {config_file_path}: {e}")
```
kaxil, line 203:
```suggestion
            logging.exception("Failed to load dag from {config_file_path}")
```

dagfactory/dagfactory.py
```diff
@@ -190,11 +190,14 @@ def load_yaml_dags(
...
     for config_file_path in candidate_dag_files:
         config_file_abs_path = str(config_file_path.absolute())
-        DagFactory(config_file_abs_path).generate_dags(globals_dict)
-        logging.info("DAG loaded: %s", config_file_path)
+        logging.info("Loading %s", config_file_abs_path)
+        try:
+            DagFactory(config_file_abs_path).generate_dags(globals_dict)
+            logging.info("DAG loaded: %s", config_file_path)
+        except Exception as e:
+            logging.exception("Failed to load dag from {config_file_path}")
```
kaxil, line 203:
```suggestion
            logging.exception("Failed to load dag from %s", config_file_path)
```

kaxil (commented):
The test needs fixing too:

```
  =========================== short test summary info ============================
  FAILED tests/test_dagfactory.py::test_load_yaml_dags_fail - Failed: DID NOT RAISE <class 'Exception'>
  ========================= 1 failed, 79 passed in 2.26s =========================

  ___________________________ test_load_yaml_dags_fail ___________________________
  
      def test_load_yaml_dags_fail():
          with pytest.raises(Exception):
  >           load_yaml_dags(
                  globals_dict=globals(),
                  dags_folder="tests/fixtures",
                  suffix=["invalid_yaml.yml"],
              )
  E           Failed: DID NOT RAISE <class 'Exception'>
```

quydx:
Hi, sorrry for late response. About the failed test case, I see the test checks if there is an exception raises when loading invalid yaml files, but this PR seems to prevent the exception so can we remove this test case `test_load_yaml_dags_fail` @tatiana ?

tatiana:
@quydx I added tests in 90487b00c1048a8fb9251fb2e9d57654b51adf36 and e9a3c728f822f0205f7cd9ae8ffc9f1b8619fe87. They are based on the PR #236 

The idea is not to delete the test but to check if we're logging the error message as expected so that this type of error isn't hidden.

Additionally, I thought it was helpful to add an invalid `yml` to the `/examples` folder and guarantee that if we run `airflow dags list-import-errors` using that folder as the dags folder, the valid symbols and DAGs would continue being imported without errors.

With these, the feature is fully covered with tests and we can guarantee the new behaviour is as intended.

tatiana (approved):

quydx:
Greate idea @tatiana , it seems more comprehensive way to handle this case. Things are good now

PR #191, opened 2024-07-17
Update the license from MIT to Apache 2.0

pankajastro:
- Update the license from MIT to Apache 2.0
- Remove Python <3.8 and Airflow < 2.0.0 Reference
- Update the project Author
- Remove FUNDING.yml

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,12 +92,12 @@ setup(
     install_requires=REQUIRED,
     extras_require={"dev": DEV_REQUIRED},
     include_package_data=True,
-    license="MIT",
+    license="Apache 2.0",
     keywords="airflow",
     classifiers=[
         # Trove classifiers
         # Full list: https://pypi.python.org/pypi?%3Aaction=list_classifiers
-        "License :: OSI Approved :: MIT License",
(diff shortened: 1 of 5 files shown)


pankajkoti (approved):

setup.py
```diff
@@ -92,15 +92,15 @@ def run(self):
...
     keywords="airflow",
     classifiers=[
         # Trove classifiers
         # Full list: https://pypi.python.org/pypi?%3Aaction=list_classifiers
-        "License :: OSI Approved :: MIT License",
+        "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.8",
```
kaxil, line 103: Since it is >3.8

We could add more versions too

kaxil (approved):

setup.py
pankajastro: For now, I have added Python3.{8,9,10}. I'll add 3.11, 3.12 once will start running tests for it

PR #202, opened 2024-07-18
Remove unnecessary config line

jlaneve:
I'm _pretty_ sure this is a mistake so removing it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,6 @@ producer_dag:
       operator: airflow.operators.bash_operator.BashOperator
       bash_command: "echo 1"
       outlets: [ 's3://bucket_example/raw/dataset1.json' ]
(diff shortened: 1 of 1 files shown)


pankajastro (approved):

cmarteepants:
Ah I see yes there is a typo. However, the `task_2` key should be there so the config matches the example in the screenshot. `[custom_operators.png](..%2F..%2FDesktop%2Fcustom_operators.png)` should not. It should be:
```
    task_2:
      bash_command: "echo 2"
      dependencies: [ task_1 ]
```

We can either reopen the PR and re-add the `task_2` key, or remove the rest of the `task_2` config and update the screenshot.

cmarteepants:
Went with the former as it was easier

Issue #194, opened 2024-07-18
Use `pyproject.toml` for the project build


pankajastro:
Acceptance criteria:

- [x]  Remove setup.py
- [x] Create pyproject.toml to build and test

PR #208, opened 2024-07-25
Use Hatchling to modernize package building

kaxil:
Replace `setup.py` with `pyproject.toml` to modernize packaging.

closes https://github.com/astronomer/dag-factory/issues/194


tatiana (approved):
Looks great, @kaxil , thank you!

PR #219, opened 2024-09-17
[WEBSITE] Add scarf to readme for website analytics

cmarteepants:
This adds website analytics to the Dag-Factory readme. Scarf privacy policy: https://about.scarf.sh/privacy-policy

Note that while you cannot explicitly opt-out of website analytics for the publicly hosted readme (and docs), Scarf respects browser DND. If that is set via the browser, telemetry for that user will not be sent to Scarf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 [![PyPi](https://img.shields.io/pypi/v/dag-factory.svg)](https://pypi.org/project/dag-factory/)
 [![Code Style](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
 [![Downloads](https://pepy.tech/badge/dag-factory)](https://pepy.tech/project/dag-factory)
+<img referrerpolicy="no-referrer-when-downgrade" src="https://static.scarf.sh/a.png?x-pxid=2bb92a5b-beb3-48cc-a722-79dda1089eda" />
 
 Welcome to *dag-factory*! *dag-factory* is a library for [Apache Airflow®](https://airflow.apache.org) to construct DAGs declaratively via configuration files. 
 
--- a/PRIVACY_NOTICE.md
+++ b/PRIVACY_NOTICE.md
@@ -1,0 +1,3 @@
+# Privacy Notice
+
+This project follows the [Privacy Policy of Astronomer](https://www.astronomer.io/privacy/)


README.md
```diff
@@ -5,6 +5,7 @@
 [![PyPi](https://img.shields.io/pypi/v/dag-factory.svg)](https://pypi.org/project/dag-factory/)
 [![Code Style](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
 [![Downloads](https://pepy.tech/badge/dag-factory)](https://pepy.tech/project/dag-factory)
+<img referrerpolicy="no-referrer-when-downgrade" src="https://static.scarf.sh/a.png?x-pxid=2bb92a5b-beb3-48cc-a722-79dda1089eda" />
 
 Welcome to *dag-factory*! *dag-factory* is a library for [Apache Airflow®](https://airflow.apache.org) to construct DAGs declaratively via configuration files.
```
tatiana, line 11: @cmarteepants, it is important that we say in our docs the notice you wrote in the PR description, referencing
> This adds website analytics to the Dag-Factory readme. Scarf privacy policy: https://about.scarf.sh/privacy-policy

Or https://airflow.apache.org/privacy-notice/ or https://privacy.apache.org/policies/privacy-policy-public.html, which the community may more know.

cmarteepants: Looking at other apache examples such as [Apache Hudi](https://github.com/apache/hudi/pull/10775/files), I don't see any specific disclaimers to website analytics for readme's. Are there any examples that you can point to where we say that tracking pixels are used in the readme?

README.md
```diff
@@ -5,6 +5,7 @@
 [![PyPi](https://img.shields.io/pypi/v/dag-factory.svg)](https://pypi.org/project/dag-factory/)
 [![Code Style](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
 [![Downloads](https://pepy.tech/badge/dag-factory)](https://pepy.tech/project/dag-factory)
+<img referrerpolicy="no-referrer-when-downgrade" src="https://static.scarf.sh/a.png?x-pxid=2bb92a5b-beb3-48cc-a722-79dda1089eda" />
```
tatiana, line 8: This seems to be rendered as
```
<img src="https://camo.githubusercontent.com/4b5c60ccaf0796d9330635a0927b42159b34946c2bf4d7aa4e412c8b1f606f59/68747470733a2f2f706570792e746563682f62616467652f6461672d666163746f7279" alt="Downloads" data-canonical-src="https://pepy.tech/badge/dag-factory" style="max-width: 100%;">
```
Probably due to:
> Pixel-based telemetry will work on standard webpages, rendered markdown documentation on package registry sites like Docker Hub, npm, and PyPi, and anywhere an image can be embedded, with a notable exception being GitHub. When GitHub renders markdown, it rewrites URLs from their original web address to https://camo.githubusercontent.com/, where GitHub hosts any linked images themselves. This prevents Scarf from providing insights to maintainers, since all that can now be detected at the original web address via the tracking pixel is undifferentiated traffic from GitHub.

Paragraph from: https://docs.scarf.sh/web-traffic/

README.md
tatiana: Apache Airflow docs: https://airflow.apache.org/privacy-notice/ .
We can have a dedicated page for this as well - it's just that ATM in DAG Factory all the doc lives in the README.

README.md
tatiana: That said, we should be able to see this change in https://pypi.org/project/dag-factory/ once we publish this change to PyPI.

cmarteepants: Confirmed with Arjun that it still works, but that it won't be as fine-grained.

README.md
cmarteepants: Added a document that redirects to Astronomer's privacy policy

README.md
tatiana: When users access the `README.md` hosted on Github, Scarff will believe that the traffic is coming from Github IPs, as if they were performing the web traffic.

This means that we won't have conversion rates from viewing docs to downloading your artifacts and possibly not which parts of your documentation are looked at most when the access comes from Github, which seem to be the main features of this:
https://docs.scarf.sh/web-traffic/

That said, the data that comes from PyPI, once the package is published should be accurate. Would it be possible for us to filter out the misleading information added by Github? Or would it make sense to have a README for PyPI that is not the same as the Github one - and only add the pixel to the PyPI README?

cmarteepants: I used the pixel that was specific for the readme. Yes, we should use a separate one for pypi. It would interesting to compare, and if we find the one directly embedded in the readme isn't useful, we can remove it at that time.

README.md
tatiana: Thanks a lot, @cmarteepants !

README.md
tatiana: Should we make the PyPI change in a separate PR?

tatiana (approved):

README.md
tatiana: For the record, it seems these docs in Scarf are outdated:
> When GitHub renders markdown, it rewrites URLs from their original web address to https://camo.githubusercontent.com/, where GitHub hosts any linked images themselves. This prevents Scarf from providing insights to maintainers, since all that can now be detected at the original web address via the tracking pixel is undifferentiated traffic from GitHub.

In practice, we currently (20 September) can track pixel-based events in Github markdown pages with the current changes. We were able to see locations and companies in the Scarf UI.

PR #234, opened 2024-10-10
Fix the duplicate test name

pankajastro:
The test with def `test_get_dag_params` is twice declare in test_dagbuilder.py. This PR update the test name.

### DIFF
--- a/tests/test_dagbuilder.py
+++ b/tests/test_dagbuilder.py
@@ -1,12 +1,11 @@
-import os
 import datetime
+import os
 from pathlib import Path
 from unittest.mock import patch
 
 import pendulum
 import pytest
 from airflow import DAG
-
 from packaging import version
 
 try:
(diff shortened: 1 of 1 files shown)


tests/test_dagbuilder.py
```diff
@@ -37,7 +36,7 @@
 from dagfactory import dagbuilder
 
 if version.parse(AIRFLOW_VERSION) >= version.parse("2.0.0"):
-    from airflow.timetables.interval import CronDataIntervalTimetable
+    pass
```
tatiana, line 39: Was this a dead code?

pankajastro: Removed it

tests/test_dagbuilder.py
```diff
@@ -461,7 +454,7 @@ def test_build():
         assert actual["dag"].tags == ["tag1", "tag2"]
 
 
-def test_get_dag_params():
+def test_get_dag_params_default_args():
```
pankajkoti, line 457:
```suggestion
def test_get_dag_params_dag_with_task_group():
```

I am not sure `default_args` if a good suffix here. The other test `test_get_dag_params` also gets `default_args`. The difference in both the tests is that the other one passes the `DAG_CONFIG` and this one passes `DAG_CONFIG_TASK_GROUP` to DagBuilder.

pankajkoti (approved):
LGTM. One minor suggestion about the naming of the test

PR #235, opened 2024-10-10
Cast callbacks to functions when set with default_args on task groups

pankajastro:
Hello!

When used in the task group's default_args, callbacks are passed as strings to the tasks, which ends up generating failures.

The idea is to perform the same parse used in make_task (it is necessary to check whether the default_args attribute is defined, since it is optional and defaults to None) and perhaps in the future modularize this conversion functionality in order to cover the widespread use of default_args.

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -179,8 +179,16 @@ class DagBuilder:
 
         if utils.check_dict_key(dag_params["default_args"], "sla_miss_callback"):
             if isinstance(dag_params["default_args"]["sla_miss_callback"], str):
-                dag_params["default_args"]["sla_miss_callback"]: Callable = import_string(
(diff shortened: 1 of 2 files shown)


tatiana (approved):
Thanks for addressing this, @pankajastro ! Should we merge #174 first and then immediately after this PR, to fix the static checks that are not passing?

pankajkoti:
It appears to me that we have not enabled the setting to update branches/PR to latest main before merging the PR. Can we please enable that?

pankajastro:
> Thanks for addressing this, @pankajastro ! Should we merge #174 first and then immediately after this PR, to fix the static checks that are not passing? Please improve the description of the PR with what we're accomplishing in this PR.

This is a replication of #17 only. I attempted to address the checks in the original PR, but it became too complicated to implement changes there

Issue #230, opened 2024-10-09
Run tests against Airflow 2.2.5 - 2.10


tatiana:
At the moment, tox is only running tests against Airflow 2.2.5.

We want to be able to run tests against all minor versions of Airflow until the latest (2.10) - making sure DAG Factory is compatible with the last releases.

PR #189, opened 2024-07-12
Update quickstart and add feature examples

cmarteepants:
This PR:
- simplifies quickstart instructions
- adds feature examples
- removes support for Python 3.7.

PR #237, opened 2024-10-11
Install Airflow with different versions in CI & run corresponding jobs

pankajkoti:
closes: #230
closes: #238 
closes: #239 
closes: #240 
closes: #242 
closes: #243
closes: #244

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,11 +22,11 @@ jobs:
   Static-Check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
-      - uses: actions/setup-python@v3
+      - uses: actions/setup-python@v5
         with:
           python-version: "3.12"
           architecture: "x64"
@@ -76,11 +76,11 @@ jobs:
           - python-version: "3.12"
             airflow-version: "2.8"
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           path: |
             ~/.cache/pip
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,7 @@ jobs:
   Run-Unit-Tests:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
         airflow-version: [ "2.2", "2.3", "2.4", "2.5", "2.6", "2.7", "2.8", "2.9", "2.10" ]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -108,3 +108,35 @@ jobs:
           name: coverage-unit-test-${{ matrix.python-version }}-${{ matrix.airflow-version }}
           path: .coverage
           include-hidden-files: true
+
+  Code-Coverage:
+    if: github.event.action != 'labeled'
+    needs:
+      - Run-Unit-Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install coverage
+        run: |
+          pip3 install coverage
+      - name: Download all coverage artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: ./coverage
+      - name: Combine coverage
+        run: |
+          coverage combine ./coverage/coverage*/.coverage
+          coverage report
+          coverage xml
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          fail_ci_if_error: true
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage.xml
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,14 +42,29 @@ jobs:
         python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
         airflow-version: [ "2.2", "2.3", "2.4", "2.5", "2.6", "2.7", "2.8", "2.9", "2.10" ]
         exclude:
+          # Apache Airflow versions prior to 2.3.0 have not been tested with Python 3.10
+          # See: https://airflow.apache.org/docs/apache-airflow/2.2.0/installation/prerequisites.html
+          - python-version: "3.10"
+            airflow-version: "2.2"
+          # Apache Airflow versions prior to 2.6.2 have not been tested with Python 3.11
+          - python-version: "3.11"
+            airflow-version: "2.2"
+          - python-version: "3.11"
+            airflow-version: "2.3"
           - python-version: "3.11"
             airflow-version: "2.4"
           - python-version: "3.11"
             airflow-version: "2.5"
+          - python-version: "3.11"
+            airflow-version: "2.6"
           # Apache Airflow versions prior to 2.9.0 have not been tested with Python 3.12.
           # Official support for Python 3.12 and the corresponding constraints.txt are available only for Apache Airflow >= 2.9.0.
           # See: https://github.com/apache/airflow/tree/2.9.0?tab=readme-ov-file#requirements
           # See: https://github.com/apache/airflow/tree/2.8.4?tab=readme-ov-file#requirements
+          - python-version: "3.12"
+            airflow-version: "2.2"
+          - python-version: "3.12"
+            airflow-version: "2.3"
           - python-version: "3.12"
             airflow-version: "2.4"
           - python-version: "3.12"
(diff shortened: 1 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -2,7 +2,7 @@ name: test
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [main,airflow-version-tests]
```
pankajkoti, line 5:
```suggestion
    branches: [main]
```
Remove before merging.

.github/workflows/test.yml
```diff
@@ -33,3 +33,78 @@ jobs:
 
       - run: pip3 install hatch
       - run: CONFIG_ROOT_DIR=`pwd`"/dags" hatch run tests.py3.12-2.10:static-check
+
+  Run-Unit-Tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
pankajkoti, line 40:
```suggestion
```
Decide if we want to keep this.

.github/workflows/test.yml
```diff
@@ -33,3 +33,78 @@ jobs:
...
+          - python-version: "3.12"
+            airflow-version: "2.5"
+          - python-version: "3.12"
+            airflow-version: "2.6"
+          - python-version: "3.12"
+            airflow-version: "2.7"
+          - python-version: "3.12"
+            airflow-version: "2.8"
+    steps:
+      - uses: actions/checkout@v3
```
pankajastro, line 79:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/test.yml
```diff
@@ -33,3 +33,78 @@ jobs:
...
+          CONFIG_ROOT_DIR=`pwd`"/dags" hatch -e tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }} run pip freeze
+
+      - name: Test DAG Factory against Airflow ${{ matrix.airflow-version }} and Python ${{ matrix.python-version }}
+        run: |
+          CONFIG_ROOT_DIR=`pwd`"/dags" hatch run tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }}:test-cov
+
+      - name: Upload coverage to Github
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-unit-test-${{ matrix.python-version }}-${{ matrix.airflow-version }}
```
pankajastro, line 108: We may need to add this repo in CodeCov https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage

.github/workflows/test.yml
```diff
@@ -33,3 +33,78 @@ jobs:
...
       - run: pip3 install hatch
       - run: CONFIG_ROOT_DIR=`pwd`"/dags" hatch run tests.py3.12-2.10:static-check
+
+  Run-Unit-Tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.8", "3.9", "3.10", "3.11", "3.12" ]
+        airflow-version: [ "2.2", "2.3", "2.4", "2.5", "2.6", "2.7", "2.8", "2.9", "2.10" ]
```
pankajastro, line 43: If we are enabling older Airflow version do we need to allow older Python version too like Python3.6?

pyproject.toml
```diff
@@ -64,17 +91,22 @@ include = ["dagfactory"]
...
+# THIRD PARTY TOOLS
+######################################
 
 [tool.black]
 line-length = 120
 target-version = ['py39', 'py310', 'py311', 'py312']
 
+[tool.isort]
+profile = "black"
+known_third_party = ["airflow", "jinja2"]
```
pankajastro, line 108: I am not sure if we are using Jinja directly in this project
```suggestion
known_third_party = ["airflow"]
```

pyproject.toml
```diff
@@ -44,13 +44,40 @@ dev = [
...
+######################################
+# TESTING
+######################################
+
 [tool.hatch.envs.tests]
 dependencies = [
-    "dag-factory[tests]"
-    ]
+    "dag-factory[tests]",
+    "apache-airflow~={matrix:airflow}.0,!=2.9.0,!=2.9.1",  # https://github.com/apache/airflow/pull/39670
```
pankajastro, line 63: this should be just >=2.2?

scripts/test/unit.sh
```diff
@@ -0,0 +1,5 @@
+pytest \
+    -vv \
+    --durations=0 \
+    -m "not (integration or perf)" \
+    --ignore=tests/test_example_dags.py
```
pankajastro, line 5:
```suggestion
    --durations=0
```
We do not have integration or perf ATM

scripts/test/unit-cov.sh
```diff
@@ -0,0 +1,8 @@
+pytest \
+    -vv \
+    --cov=dagfactory \
+    --cov-report=term-missing \
+    --cov-report=xml \
+    --durations=0 \
+    -m "not (integration or perf)" \
+    --ignore=tests/test_example_dags.py
```
pankajastro, line 8:
```suggestion
    --durations=0
```
We do not have integration or perf ATM

pyproject.toml
```diff
@@ -44,13 +44,40 @@ dev = [
     "tox",
 ]
 tests = [
+    "pytest>=6.0",
+    "pytest-split",
+    "pytest-dotenv",
```
pankajastro, line 49: I'm not sure if we use these lib

pyproject.toml
```diff
@@ -44,13 +44,40 @@ dev = [
...
+    "requests-mock",
+    "pytest-cov",
+    "pytest-describe",
```
pankajastro, line 52:
```suggestion
    "pytest-cov",
```
same here. I am not sure if we need

pyproject.toml
```diff
@@ -64,17 +91,22 @@ include = ["dagfactory"]
...
-static-check = " pre-commit run --files dagfactory/*"
+######################################
+# THIRD PARTY TOOLS
+######################################
 
 [tool.black]
 line-length = 120
 target-version = ['py39', 'py310', 'py311', 'py312']
 
+[tool.isort]
```
pankajastro, line 106: isort pre-commit hook is not in our pre-commit config file at the moment if we are adding this configuration, I think we should add the hook also in https://github.com/astronomer/dag-factory/blob/main/.pre-commit-config.yaml

pankajastro:
Shall we clean the makefile like removing the test target?

pyproject.toml
pankajkoti: test matrix values are accepted and used here for corresponding Airflow version environments using hatch.

.github/workflows/test.yml
pankajkoti: I think we should not since Python 3.6 is already EOL. same for 3.7. 3.8 is EOL too, so wondering if we should remove that too :)

.github/workflows/test.yml
pankajkoti: yes, @tatiana could you please help here when possible?

pyproject.toml
pankajkoti: removed. thanks.

pyproject.toml
pankajkoti: removed. thanks.

pyproject.toml
pankajkoti: made it as part of tool.ruff.lint instead in the pyproject.toml so that it gets checked via the ruff pre-commit hook

pankajkoti:
> Shall we clean the makefile like removing the test target?

yes perhaps we could have a follow-up PR for the cleanup. I also feel that we do not need the `build` pipeline that we have currently, so could clean up that too

.github/workflows/test.yml
pankajkoti: It appears that we already have a codecov token configured for the repo as I can see that the codecov job in CI is able to push to Codecov. Wondering now how we can make the CodeCov reports available on the PRs.

.github/workflows/test.yml
tatiana: I think it's usually useful to see all failures, so it seems a good idea to keep this line. WDYT?

.github/workflows/test.yml
tatiana: The last release of DAG FActory (0.19.0) before the donation to Astronomer already only supported Python 2.7 and higher:
Screenshot 2024-10-15 at 10 08 21

So it looks good - we're trying to keep it compatible with previous versions.

.github/workflows/test.yml
tatiana: The token was from 5 years ago and not associated with Astronomer's account. I've added an Astronomer token, and i'm configuring the actions

tatiana: Okay, I think I configured CodeCov 🎉 
Screenshot 2024-10-15 at 10 23 38

tatiana (approved):
This looks great, thanks, @pankajkoti ! It will help us making sure we're not introducing breaking changes when we improve DAG Factory.

.github/workflows/test.yml
pankajkoti: yes, I feel keeping it helps in the developer journey :)

.github/workflows/test.yml
pankajkoti: It appears we decided to drop 3.7 in https://github.com/astronomer/dag-factory/pull/189 and continued doing so when migrated to use hatchling https://github.com/astronomer/dag-factory/pull/208. IMO, it would be wise if we do not add back 3.7. But please let me know if you feel otherwise.

pankajkoti: I am resolving this conversation as the PR needs all conversations to be resolved before merging, but please leave a comment on the PR if you've an opinion here, thanks.

Issue #206, opened 2024-07-24
Increase project quality


pankajkoti:
Description co-authored by @tatiana @pankajastro

Scope:
- [x] https://github.com/astronomer/dag-factory/issues/199
- [x] https://github.com/astronomer/dag-factory/issues/198
- [x] https://github.com/astronomer/dag-factory/issues/196
- [x] https://github.com/astronomer/dag-factory/issues/197
- [x] #230

matveykortsev:
Guys add please support for on_failure_callback in this way:
```
from airflow.providers.slack.notifications.slack import send_slack_notification
'on_failure_callback': [
        send_slack_notification(
            slack_conn_id='slack',
            text="""
                :red_circle: Task Failed. 
                *Task*: {{ ti.task_id }}  
                *Dag*: {{ ti.dag_id }} 
                *Execution Time*: {{ ti.execution_date }}  
                *Log Url*: {{ ti.log_url }} 
                """,
            channel="analytics-alerts",
            username="Airflow",
        )
    ],
```
Now its only possible without params and only through absolute path for .py file.

pankajastro:
Thank you, @matveykortsev, indeed, this feature would be valuable. I've created a high-level task to compare the missing issues like this.

PR #249, opened 2024-10-15
Fix `make docker-run`

pankajkoti:
The `make docker-run` currently appears to be broken. 
The changes in this PR attempt to resolve the errors by installing Airflow
with its constraints. 

related: #206


tatiana (approved):
Thanks for fixing this, @pankajkoti !

Issue #214, opened 2024-08-07
[Feature] Add Scarf-based Telemetry to DAG-Factory


cmarteepants:
## Description
Astronomer is standardizing on Scarf for telemetry for our open source projects. We have an account [here](https://app.scarf.sh/home/astronomer).

Similar to https://github.com/apache/airflow/pull/39510, we would like to integrate with Scarf to start gathering telemetry to gain insight into adoption trends and feature usage. I would like to release an mvp in Q3.

Next Steps:

- [ ]  Define metrics for POC / MVP
- [ ]  Create POC and figure out which release to include this in

tatiana:
I had a meeting about this with @cmarteepants and Scarf team members yesterday.
- I now have access to Astronomer's Scarf account
- We will be creating a file-based Scarf project for DAG Factory: https://app.scarf.sh/new-package/astronomer
- More information on the approach that will be used: https://docs.scarf.sh/custom-telemetry/
- We will add Scarf analysis in the DAG Factory docs:

Things we may want to answer:
- Version of Airflow
- Version of Scarf
- How many DAGs are being created with DAG Factory
- How many DAG Runs happened

I'll add more details as we progress.

tatiana:
@cmarteepants did the first step on this, by adding a tracking pixel to our README:
https://github.com/astronomer/dag-factory/pull/219

PR #250, opened 2024-10-15
Support telemetry using Scarf

tatiana:
Export telemetry related to DAG Factory usage to [Scarf](https://about.scarf.sh/).

This data assists the project maintainers in better understanding how DAG Factory is used. Insights from this telemetry are critical for prioritizing patches, minor releases, and security fixes. Additionally, this information supports critical decisions related to the development road map.

Deployments and individual users can opt out of analytics by setting the configuration:

```
[dag_factory] enable_telemetry False
```

As described in the [official documentation](https://docs.scarf.sh/gateway/#do-not-track), it is also possible to opt-out by setting one of the following environment variables:

```commandline
AIRFLOW__DAG_FACTORY__ENABLE_TELEMETRY=False
DO_NOT_TRACK=True
SCARF_NO_ANALYTICS=True
```

In addition to Scarf's default data collection, DAG Factory collects the following information:

- DAG Factory version
- Airflow version
- Python version
- Operating system & machine architecture
- Event type
- Number of DAGs
- Number of TaskGroups
- Number of Tasks

No user-identifiable information (IP included) is stored in Scarf, even though Scarf infers information from the IP, such as location, and stores that. The data collection is GDPR compliant.

The data is not currently being emitted for pre-releases except from integration tests.

The Apache Foundation supports this same strategy in many of its OpenSource projects, including Airflow ([#39510](https://github.com/apache/airflow/pull/39510)).

Example of visualisation of the data via the Scarf UI:

Screenshot 2024-10-17 at 01 56 09

Screenshot 2024-10-17 at 01 55 59

Screenshot 2024-10-17 at 01 55 47

Screenshot 2024-10-17 at 01 55 42

Closes: #214


dagfactory/dagfactory.py
```diff
@@ -106,9 +111,20 @@ def build_dags(self) -> Dict[str, DAG]:
                 dags[dag["dag_id"]]: DAG = dag["dag"]
             except Exception as err:
                 raise DagFactoryException(f"Failed to generate dag {dag_name}. verify config is correct") from err
+            else:
+                self.taskgroups_count = dag_builder.taskgroups_count
```
pankajkoti, line 115: I am wondering what would the count represent here. Is it the count of all the taskgroups across all the DAGs in the deployment? If so, do we need to keep incrementing this count for each of the DAG?

dagfactory/dagfactory.py
```diff
@@ -106,9 +111,20 @@ def build_dags(self) -> Dict[str, DAG]:
...
+                self.tasks_count = dag_builder.tasks_count
```
pankajkoti, line 116: Same question as above for the taskgroups count.

pankajkoti (approved):
Looks great to me. Just have minor questions inline about the taskgroups and tasks count.

dagfactory/dagfactory.py
tatiana: Good catch, @pankajkoti , please, take a look at ad1063af6fc97a81d4dd8799ac51dc549a0dc9c8 - I think I addressed the issue and also added a test case to cover the scenario of a YAML file that would contain multiple DAGs.

dagfactory/dagfactory.py
tatiana: Addressed in https://github.com/astronomer/dag-factory/pull/250#discussion_r1804325449

dagfactory/dagfactory.py
pankajkoti: Thanks Tatiana. Looks great and my questions are answered

PR #255, opened 2024-10-17
Configure GitHub to automate publishing DAG Factory in PyPI

tatiana:
Automate the release process using GitHub actions so we can consistently make release 0.20.0 and future releases.

As part of this change, I've already generated a PyPI upload token and configured the environment variable `PYPI_TOKEN` as a GitHub secret.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,8 +7,8 @@
       - Code-Coverage
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: '3.10'
           architecture: 'x64'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,43 @@
+name: Publish package in PyPi
+
+on:
+  release:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,20 @@
...
+    if: github.event_name == 'release'
+    name: Build and publish Python 🐍 distributions 📦 to PyPI
+    needs:
+      - Static-Check
+      - Run-Unit-Tests
+      - Code-Coverage
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
```
pankajastro, line 11:
```suggestion
      - uses: actions/checkout@v4
      - uses: actions/setup-python@v4
```

pankajastro (approved):

.github/workflows/release.yml
pankajkoti: setup-python has v5 available

pankajkoti (approved):

.github/workflows/release.yml
tatiana: Thanks a lot, @pankajastro and @pankajkoti  - I addressed the feedback in 667d34309a96c8d6b73190d53e45f9fe860444ad

Issue #199, opened 2024-07-18
Improve code coverage


pankajastro:
Acceptance criteria:

- [ ] Code coverage >= 90%

PR #258, opened 2024-10-17
Add more tests

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/199

- Add test for the outlets param and load YML dag with default suffix cases
- Cleanup some import for Airflow < 2.0.0


tests/test_dagfactory.py
```diff
@@ -429,7 +429,15 @@ def test_load_yaml_dags_succeed(mock_emit_usage_metrics_if_enabled):
...
     args = mock_emit_usage_metrics_if_enabled.call_args.args
     assert args[0] == "load_yaml_dags"
     assert args[1] == {"dags_count": 2, "tasks_count": 4, "taskgroups_count": 0}
+    
+    # test load with none suffix
+    load_yaml_dags(
+        globals_dict=globals(),
+        dags_folder="tests/fixtures",
+    )
+
```
tatiana, line 443: It would be great to have a dedicated test for this use case, having asserts that confirm the behaviour is as expected.

tatiana (approved):
Thanks, @pankajastro , a minor comment, and we're good to merge this!

tests/test_dagfactory.py
pankajastro: Moved this into a new test and added assert to check the log

PR #263, opened 2024-10-18
fix: remove duplicated operator

brair:


tatiana (approved):
Thanks for fixing this, @brair air!

Issue #198, opened 2024-07-18
Add integration tests


pankajastro:
Acceptance criteria:

- [ ] Run integration tests in CI

PR #266, opened 2024-10-18
Run integration tests in CI

pankajkoti:
closes: #198 
closes: #206

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -2,7 +2,7 @@ name: Test and release related jobs
 
 on:
   push: # Run on pushes to the default branch
-    branches: [main]
+    branches: [integration-tests]
   pull_request_target: # Also run on pull requests originated from forks
     branches: [main]
   release:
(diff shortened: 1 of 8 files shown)


.github/workflows/cicd.yaml
```diff
@@ -104,10 +104,110 @@ jobs:
...
+
+      - name: Install packages and dependencies
+        run: |
+          python -m pip install uv
+          uv pip install --system hatch
+          hatch -e tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }} run pip freeze
+
+      - name: Test DAG Factory against Airflow ${{ matrix.airflow-version }} and Python ${{ matrix.python-version }}
+        run: |
+          PYTHONPATH=`pwd`:`pwd`/examples CONFIG_ROOT_DIR=`pwd`"/dags" hatch run tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }}:test-integration-setup
```
tatiana, line 185: Would it make sense to set
```
PYTHONPATH: $PWD:$PWD/examples
```
Or
```
PYTHONPATH: .:./examples
```

In the `env` section (currently line 187)? Same for `CONFIG_ROOT_DIR`

tatiana (approved):
This looks great, @pankajkoti , really happy we'll have these integration tests to give us more confidence when making changes and accepting contributions to DAG factory.

I left a minor feedback inline.

.github/workflows/cicd.yaml
```diff
@@ -104,10 +104,110 @@ jobs:
...
+            airflow-version: "2.4"
+          - python-version: "3.12"
+            airflow-version: "2.5"
+          - python-version: "3.12"
+            airflow-version: "2.6"
+          - python-version: "3.12"
+            airflow-version: "2.7"
+          - python-version: "3.12"
+            airflow-version: "2.8"
+    services:
```
pankajastro, line 148: Does our test use this Postgres sevice?

pankajastro (approved):

.github/workflows/cicd.yaml
pankajkoti: yes, thanks. setting & using env vars now. $PWD din't seem to evaluate well. I used ${{ github.workflow }} and that seems to work well.

.github/workflows/cicd.yaml
pankajkoti: good check, yes, we don't need postgres service, cleaned it up :)

Issue #193, opened 2024-07-18
Remove legacy `Dockerfile `


pankajastro:
Acceptance criteria:

- [x] Remove file https://github.com/astronomer/dag-factory/blob/0375c76de5c4e37269282133caa219a9698b19ae/Dockerfile
- [x] Use astro or docker composer for docker based local build

PR #267, opened 2024-10-18
Add astro cli project to run airflow+dagfactory

pankajastro:
Summary: This PR introduces the Astro CLI project into the repository and removes the legacy Dockerfile.

Changes:

- **Added:** Astro CLI project to facilitate managing the Airflow environment.
- **Removed:** Legacy Dockerfile to streamline the project and avoid confusion.
- **Usage:** To start the Airflow container, run the following command:
```shell
make docker-run
``` 

closes: https://github.com/astronomer/dag-factory/issues/193


dev/.astro/dag_integrity_exceptions.txt
```diff
@@ -0,0 +1 @@
+# Add dag files to exempt from parse test below. ex: dags/<test-file>
```
tatiana, line 1: We probably should delete this if we're not using

dev/.astro/test_dag_integrity_default.py
```diff
@@ -0,0 +1,130 @@
+"""Test the validity of all DAGs. **USED BY DEV PARSE COMMAND DO NOT EDIT**"""
```
tatiana, line 1: We probably should delete this if we're not using

dev/tests/dags/test_dag_example.py
```diff
@@ -0,0 +1,74 @@
+"""Example DAGs test. This test ensures that all Dags have tags, retries set to two, and no import errors. This is an example pytest and may not be fit the context of your DAGs. Feel free to add and remove tests."""
```
tatiana, line 1: We probably should delete this if we're not using

Makefile
```diff
@@ -27,16 +27,24 @@ clean: ## Removes build and test artifacts
...
-	@docker build -t dag_factory:latest .
+
+.PHONY: copy-files
+copy-files: ## copy project in docker context
+	mkdir -p dev/include/tmp_dagfactory
+	cp -r ./dagfactory dev/tmp_dagfactory
+	cp ./pyproject.toml dev/include/tmp_dagfactory/pyproject.toml
+	cp ./LICENSE dev/include/tmp_dagfactory/LICENSE
+	cp ./README.md dev/include/tmp_dagfactory/README.md
+	cp -r ./examples/* dev/dags/
```
tatiana, line 38: Why don't we just move examples to `dev/dags`?

Makefile
```diff
@@ -27,16 +27,24 @@ clean: ## Removes build and test artifacts
...
-	@echo "==> Building docker image for local testing"
-	@docker build -t dag_factory:latest .
+
+.PHONY: copy-files
+copy-files: ## copy project in docker context
+	mkdir -p dev/include/tmp_dagfactory
+	cp -r ./dagfactory dev/tmp_dagfactory
+	cp ./pyproject.toml dev/include/tmp_dagfactory/pyproject.toml
+	cp ./LICENSE dev/include/tmp_dagfactory/LICENSE
+	cp ./README.md dev/include/tmp_dagfactory/README.md
```
tatiana, line 37: Why are we copying all these files? Would there be an alternative (e.g. to mount `dagfactory` in a folder that is part of the PYTHONPATH in the `Dockerfile`? Or some other..?

dev/tests/dags/test_dag_example.py
pankajastro: you deploy on Astro Cloud using astro deploy, it helps identify any import errors in DAG files early in the process. However, please let me know if you think it would be better to remove it.

Makefile
pankajastro: We could move the example in dags files, but maybe we need to adjust paths in some tests/readme, etc. Please do let me know if we want to do this changes in this PR?

Makefile
pankajastro: > Why are we copying all these files? Would

We need these files to install dagfactory local changes in astro project.

> Would there be an alternative (e.g. to mount dagfactory in a folder that is part of the PYTHONPATH in the Dockerfile? Or some other..?

I attempted using a symbolic link, but it didn’t work. I can look into a mounting approach, though I'm not sure at the moment if it will work or not.

dev/tests/dags/test_dag_example.py
tatiana: Let's keep it then, @pankajastro ..!

Makefile
tatiana: We agreed to have a symbolic link from `examples` and keep the data in `dev/dags`

Makefile
tatiana: An alternative could be to build a wheel, and install using it - this way we'll avoid copying all the files

Makefile
pankajastro: I created a symbolic link between examples/* and dev/dags but DAG is not visible in Airflow 

**Symbolic link created locally**

Screenshot 2024-10-21 at 6 52 05 PM

**Symbolic link visible in docker**

Screenshot 2024-10-21 at 6 53 21 PM

**DAG is not visible in Airflow UI**

Screenshot 2024-10-21 at 6 53 32 PM

dev/.astro/dag_integrity_exceptions.txt
pankajastro: this is getting used in dev/.astro/test_dag_integrity_default.py

dev/.astro/test_dag_integrity_default.py
pankajastro: this also has some tests that may be helpful to catch errors before deploying

Makefile
pankajastro: The files are in the examples folder. Are you suggesting moving them to dev/dags, or am I missing something?

tatiana: Yes, @pankajastro, that's the suggestion
`git mv examples dev/dags`

And then create a symbolic link `examples`

tatiana (approved):
Thanks for addressing all the feedback, @pankajastro , this is invaluable for us to validate DAG factory in Astro

PR #269, opened 2024-10-22
Add GitHub issue template for bug reports and feature requests

pankajkoti:
Issue templates to help triage them faster.

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-bug.yml
+++ b/.github/ISSUE_TEMPLATE/01-bug.yml
@@ -8,26 +8,17 @@ body:
     attributes:
       value: |
         Thanks for taking the time to fill out this bug report!
-  - type: dropdown
+  - type: input
     id: dag-factory-version
     attributes:
-      label: DAG Factory Version
-      description: What version of DAG Factory are you running? If you do not see your version in the list, please (ideally) test on
-        the latest release or main to see if the issue is fixed before reporting it.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/01-bug.yml
```diff
@@ -0,0 +1,123 @@
...
+  - type: dropdown
+    id: dag-factory-version
+    attributes:
+      label: DAG Factory Version
+      description: What version of DAG Factory are you running? If you do not see your version in the list, please (ideally) test on
+        the latest release or main to see if the issue is fixed before reporting it.
+      options:
+        - "0.19.0"
+        - "main (development)"
+        - "Other DAG Factory version (please specify below)"
```
tatiana, line 20: We should probably add an example but not hard-coded versions. It probably does not make sense for us to have to update this every release we do - it will become stale unless we find a way to automate this.

tatiana (commented):
It will be great to have some templates, @pankajkoti , thanks for creating these! Minor feedback inline

.github/ISSUE_TEMPLATE/01-bug.yml
pankajkoti: yes, make sense totally! Removed the dropdown and we now have a free text input to accept the version.

tatiana (approved):

PR #270, opened 2024-10-22
Add CODEOWNERS

pankajkoti:
Add CODEOWNERS so that they are automatically requested for review when someone opens a pull request.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+@pankajastro @pankajkoti @tatiana
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-@pankajastro @pankajkoti @tatiana
+@oss-integrations


CODEOWNERS
```diff
@@ -0,0 +1 @@
+@pankajastro @pankajkoti @tatiana
```
pankajkoti, line 1:
```suggestion
@astronomer/astro-dagfactory-codeowners
```

Check instead if such an alias/group with adding those team members can be created & used here

tatiana: I just asked @phanikumv for us to have a single GH team for the OSS Integrations team, and then we could assign this both here and in the other repos. To maintain a team per project would be just additional maintenance overhead

pankajkoti: thanks @tatiana . @phanikumv could you please let me know once you create that team and the name of the team so I can add it here and in other repos, too?

phanikumv: Added `oss-integrations` github team 

https://github.com/orgs/astronomer/teams/oss-integrations/members

pankajkoti, line 1:
```suggestion
@oss-integrations
```

pankajkoti: Thank you @phanikumv. Applied the change.

tatiana (approved):
Looks great, thanks a lot, @pankajkoti !

PR #273, opened 2024-10-22
Remove off looking start-date value in example_dag yaml config

pankajkoti:
The `example_dag` in the `example_dag_factory.yml` has a `start_date` 
value set to `2 days`, which seems unclear and problematic. It’s not clear 
what this value represents, and it appears to be causing issues with the 
DAG, preventing it from running tasks. I confirmed this issue also exists in 
version `0.19.0`, so it’s not related to any changes made after that version. 
When I removed this questionable `start_date` value, the DAG ran its tasks 
successfully.

closes: #274

### DIFF
--- a/dev/dags/example_dag_factory.yml
+++ b/dev/dags/example_dag_factory.yml
@@ -19,7 +19,6 @@ default:
 example_dag:
   default_args:
     owner: "custom_owner"
-    start_date: 2 days
   description: "this is an example dag"
   schedule_interval: "0 3 * * *"
   render_template_as_native_obj: True


pankajastro (approved):

pankajastro (commented):
shall we remove `end_date: 2018-03-05` too?

tatiana (approved):

pankajkoti:
> shall we remove `end_date: 2018-03-05` too?

I guess we can keep it, it does not harm us,  DAG runs are created & run fine until that end_date. And since we have been having it as example like this, I guess good to keep it as is without making much changes. WDYT?

pankajkoti:
> > shall we remove `end_date: 2018-03-05` too?
> 
> I guess we can keep it, it does not harm us, DAG runs are created & run fine until that end_date. And since we have been having it as example like this, I guess good to keep it as is without making much changes. WDYT?

As discussed over Slack, I feel it's good to have that start_date - end_date time window to demonstrate that we're able to run DAGs using DAG factory (accepting those params well), so I'm going ahead with the merge. If the need is to support manual DAG runs, I suggest we create & add another example DAG without having an end_date, but for the given example I strongly believe it's a good example to have an end_date. I am going ahead with the merge now. Thanks for the review & the question :)

PR #276, opened 2024-10-25
Update CODEOWNERS to track all files

pankajkoti:
The first argument in the CODEOWNERS file tells what all files to monitor for changes. Since, we din't specify any file, the code review requests are not getting sent. Specifying the wild card asterisk tells it to monitor for all files in the repo & accordingly work for the CODEOWNERS features such as sending review requests to the CODEOWNERS upon opening pull requests.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-@oss-integrations
+* @oss-integrations
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @oss-integrations
+* @astronomer/oss-integrations


CODEOWNERS
```diff
@@ -1 +1 @@
-@oss-integrations
+* @oss-integrations
```
pankajastro, line 1: look like @oss-integrations does not have access

pankajkoti: added the prefix `astronomer/` which was needed for the team

tatiana (approved):

CODEOWNERS
```diff
@@ -1 +1 @@
-@oss-integrations
+* @astronomer/oss-integrations
```
tatiana, line 1: Thanks for fixing this, @pankajkoti ! I hadn't realised before we were missing the org.

PR #281, opened 2024-11-05
Enable secrect detection precommit check

pankajastro:


pankajkoti (approved):
Thanks for adding this!

Issue #282, opened 2024-11-06
[telemetry] Identify a better strategy of when to emit telemetry


tatiana:
As of DAG Factory 0.20, we emit user activity events every time a DAG Factory DAG is parsed, as part of #250.

Tracking when a DAG is parsed does not seem to provide much value. Would there be better ways to track usage if the DAG factory was installed in remote hosts?

One possibility, for instance, is to use Airflow listeners and only emit events before or after a DAG Factory-build DAG is run: https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/listeners.html.

The goal of this ticket is to brainstorm possibilities and improve our current approach.

pankajastro:
> Tracking when a DAG is parsed doesn't seem to provide much value.

Currently, we collect `dags_count`, `tasks_count`, and `taskgroups_count`. I wanted to understand what additional data might help us. Are we looking for runtime data like `DagRun`? Could you please elaborate a bit?

I explored the Airflow listener, and I think if information like `DagRun` would be helpful, we could consider implementing a custom solution using `on_dag_run_running` listener to emit an event when a DagRun belongs to a DagFactory DAG. We could see if we can attach some parameters to the DAG to accurately identify if the DAG is from DagFactory

tatiana:
Currently, whenever these things are currently called:
- DagBuilder
- load_yaml_dags

We emit the events. (dags_count, task_count)

We emit this information every time these things are called during DAG parsing (let's say every 30 seconds).

From a telemetry perspective, the data we're collecting is proportional to the number of times the DAG is parsed. But from a business perspective, this is not so meaningful.

Some examples of what would be more valuable business metrics:
1. In an Airflow deployment: how many dags and tasks are created with DAG Factory (once, per deployment)
2. How many DAG runs are from DAGs created with DAG Factory (successful / failure)
3. How many task runs are from tasks created via DAG Factory (successful / failure)

With the listeners' approach, we could collect/emit (2) and (3) by installing the plugin:
https://github.com/astronomer/astronomer-cosmos/blob/1abb37145fcc3b756c513a60f04a659491d84761/pyproject.toml#L104-L108

We may not want to worry about (1) as a starting point - by replacing the current DAG-parsing-focused approach with the listeners' approach, we may already have much more valuable information.

PR #300, opened 2024-11-26
Refactor telemetry to collect events during DAG run and not during DAG parsing

pankajastro:
DAG Factory 0.20 started collecting telemetry as part of the PR https://github.com/astronomer/dag-factory/pull/250. However, one limitation of this initial implementation is that it emitted telemetry every time DAGs were parsed. This means that the data collected did not represent the actual usage and was proportional to the number of times a DAG was parsed. This PR aims to address this limitation by changing DAG Factory to emit telemetry during DAG runs.

This implementation leverages Airflow listeners to only emit events after a Factory-Built DAG is run.

Closes: https://github.com/astronomer/dag-factory/issues/282

With this data, we can get the following insight
- Number of failed DagRuns
- Number of successful DagRuns
- Total tasks associated to each DagRun
- DagRun hash

**Airflow Version**
Screenshot 2024-12-03 at 8 00 14 PM

**DAG Hash**
Screenshot 2024-12-03 at 8 01 28 PM

**DAG Factory Version**
Screenshot 2024-12-03 at 8 02 08 PM

**Event Type**
Screenshot 2024-12-03 at 8 02 46 PM

**Platform Machine**
Screenshot 2024-12-03 at 8 03 18 PM

**Platform System**
Screenshot 2024-12-03 at 8 04 23 PM

**Python Version**
Screenshot 2024-12-03 at 8 05 05 PM

**DAG Run Status**
Screenshot 2024-12-03 at 8 06 45 PM

**Task Count in DAG run**
Screenshot 2024-12-03 at 8 07 25 PM

** Telemetry Version**
Screenshot 2024-12-03 at 8 07 58 PM

### DIFF
--- a/dagfactory/constants.py
+++ b/dagfactory/constants.py
@@ -1,3 +1,3 @@
-TELEMETRY_URL = "https://astro.gateway.scarf.sh/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}"
+TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}"
 TELEMETRY_VERSION = "v1"
 TELEMETRY_TIMEOUT = 5.0
--- a/dagfactory/constants.py
+++ b/dagfactory/constants.py
@@ -1,3 +1,3 @@
-TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/dag-factory/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{type}/{status}/{dag_hash}/{task_count}"
+TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/dag-factory/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}"
 TELEMETRY_VERSION = "v2"
 TELEMETRY_TIMEOUT = 5.0
--- a/dagfactory/constants.py
+++ b/dagfactory/constants.py
@@ -1,3 +1,3 @@
-TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/dag-factory/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}"
+TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/dag-factory/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{type}/{status}/{dag_hash}/{task_count}"
 TELEMETRY_VERSION = "v2"
 TELEMETRY_TIMEOUT = 5.0
--- a/dagfactory/constants.py
+++ b/dagfactory/constants.py
@@ -1,3 +1,3 @@
-TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/dag-factory/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}?{query_string}"
+TELEMETRY_URL = "https://astro.gateway.scarf.sh/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}"
 TELEMETRY_VERSION = "v1"
 TELEMETRY_TIMEOUT = 5.0
--- a/dagfactory/constants.py
+++ b/dagfactory/constants.py
@@ -1,3 +1,3 @@
-TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}"
+TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/dag-factory/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}?{query_string}"
 TELEMETRY_VERSION = "v1"
 TELEMETRY_TIMEOUT = 5.0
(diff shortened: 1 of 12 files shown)


dagfactory/dagbuilder.py
```diff
@@ -795,9 +795,9 @@ def build(self) -> Dict[str, Union[str, DAG]]:
             )
             dag.doc_md = doc_md_callable(**dag_params.get("doc_md_python_arguments", {}))
 
-        # tags parameter introduced in Airflow 1.10.8
-        if version.parse(AIRFLOW_VERSION) >= version.parse("1.10.8"):
-            dag.tags = dag_params.get("tags", None)
+        tags = dag_params.get("tags", [])
+        tags.append("dagfactory")
```
tatiana, line 799: Is there any chance `dagfactory` was already added by the end-user? Can we have duplicated tags?

dagfactory/dagbuilder.py
```diff
@@ -795,9 +795,9 @@ def build(self) -> Dict[str, Union[str, DAG]]:
             )
             dag.doc_md = doc_md_callable(**dag_params.get("doc_md_python_arguments", {}))
 
-        # tags parameter introduced in Airflow 1.10.8
-        if version.parse(AIRFLOW_VERSION) >= version.parse("1.10.8"):
```
tatiana, line 799: It's always good to see this cleanup 🙌

dagfactory/listeners/runtime_event.py
```diff
@@ -0,0 +1,102 @@
...
+    def on_task_instance_failed(previous_state: TaskInstanceState, task_instance: TaskInstance, session):
+        if not is_dagfactory_dag(task_instance.dag_run.get_dag()):
+            return
+        additional_telemetry_metrics = {
+            "dag_hash": task_instance.dag_run.dag_hash,
+            "status": EventStatus.FAILED,
+            "operator": task_instance.operator,
+        }
+
+        telemetry.emit_usage_metrics_if_enabled(EventType.TASK_INSTANCE, additional_telemetry_metrics)
```
tatiana, line 102: I think emitting per DAG Run may be good enough. Otherwise, we may have a too large amount of events without having proportional value. It's probably worth checking with @cmarteepants

dagfactory/listeners/runtime_event.py
```diff
@@ -0,0 +1,102 @@
...
+        return True
+    return False
+
+
+@hookimpl
+def on_dag_run_success(dag_run: DagRun, msg: str):
+    if not is_dagfactory_dag(dag_run.get_dag()):
+        return
+    additional_telemetry_metrics = {
+        "dag_hash": dag_run.dag_hash,
```
tatiana, line 42: What does the `dag_hash` contain? May it contain any personal/company-identifiable data?

pankajastro: This is an md5 hash, so this will just be a 32bit char and not any data example hash: d151d1fa2f03270ea116cc7494f2c591

dagfactory/dagbuilder.py
pankajastro: Added a check

dagfactory/listeners/runtime_event.py
tatiana: Regarding the reversibility of MD5:
https://md5.gromweb.com/

I don't think that would be an issue. Still, we could consider r using another way of generating hashes, like using more robust algorithms ([SHA-2](https://en.wikipedia.org/wiki/SHA-2), [Whirlpool](https://en.wikipedia.org/wiki/Whirlpool_%28cryptography%29), etc), combining algorithms, and using a "[salt](https://en.wikipedia.org/wiki/Salt_%28cryptography%29)".

dagfactory/listeners/runtime_event.py
```diff
@@ -0,0 +1,99 @@
...
+    @hookimpl
+    def on_task_instance_failed(previous_state: TaskInstanceState, task_instance: TaskInstance, session):
+        if not is_dagfactory_dag(task_instance.dag_run.get_dag()):
+            return
+        additional_telemetry_metrics = {
+            "dag_hash": task_instance.dag_run.dag_hash,
+            "status": EventStatus.FAILED,
+        }
+
+        telemetry.emit_usage_metrics_if_enabled(EventType.TASK_INSTANCE, additional_telemetry_metrics)
```
tatiana, line 99: @pankajastro, as we discussed in the call earlier today, I believe having an event per task run would provide too many events without necessarily giving significant business value.

Ideally, we'd have @cmarteepants' confirmation on this, but I take the risk - let's remove it.

We can add this in the future if she believes we should re-add this.

pankajastro: Removed the TI listeners

dagfactory/listeners/runtime_event.py
pankajastro: Yes, the salt-based hash is more secure. But since Airflow calculates this hash, I think we can keep it for now that way we do not have to maintain it. but please let me know if you want me to change to some other hashing algorithm

tests/test_telemetry.py
```diff
@@ -52,15 +51,19 @@ def test_emit_usage_metrics_fails(mock_httpx_get, caplog):
...
-        "variables": {"a": 1, "b": 2},
+        "event_type": "dag_run",
+        "status": "success",
+        "dag_hash": "d151d1fa2f03270ea116cc7494f2c591",
+        "task_count": 3,
     }
     is_success = telemetry.emit_usage_metrics(sample_metrics)
     mock_httpx_get.assert_called_once_with(
-        "https://astronomer.gateway.scarf.sh/dag-factory/v1/0.2.0a1/2.10.1/3.11/darwin/amd64?dagfactory_version=0.2.0a1&airflow_version=2.10.1&python_version=3.11&platform_system=darwin&platform_machine=amd64&variables=%7B%27a%27%3A+1%2C+%27b%27%3A+2%7D",
+        "https://astronomer.gateway.scarf.sh/v1/0.2.0a1/2.10.1/3.11/darwin/amd64/dag_run/success/d151d1fa2f03270ea116cc7494f2c591/3",
```
tatiana, line 61: Why is the URL missing `/dag-factory/`?

Since Astronomer is building our own dashboards, and very few people have access to Scarf right now, wouldn't it be better to continue passing most arguments as query strings - so we can easily extend them, if needed?

pankajastro: > Why is the URL missing /dag-factory/?

Maybe I'll have to add that in path 

Screenshot 2024-11-29 at 10 36 28 PM

pankajastro: Fixed in c9f9b7104871ae619c19d2e68c97f344ee963ce7

dagfactory/constants.py
```diff
@@ -1,3 +1,3 @@
-TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/dag-factory/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}?{query_string}"
+TELEMETRY_URL = "https://astronomer.gateway.scarf.sh/dag-factory/{telemetry_version}/{dagfactory_version}/{airflow_version}/{python_version}/{platform_system}/{platform_machine}/{event_type}/{status}/{dag_hash}/{task_count}?{query_string}"
 TELEMETRY_VERSION = "v1"
```
tatiana, line 2:
```suggestion
TELEMETRY_VERSION = "v2"
```

tests/test_telemetry.py
```diff
@@ -52,15 +52,20 @@ def test_emit_usage_metrics_fails(mock_httpx_get, caplog):
         "python_version": "3.11",
         "platform_system": "darwin",
         "platform_machine": "amd64",
-        "variables": {"a": 1, "b": 2},
+        "type": "dag_run",
+        "status": "success",
+        "dag_hash": "d151d1fa2f03270ea116cc7494f2c591",
+        "task_count": 3,
     }
+    # query_string = urlencode(sample_metrics)
```
pankajastro, line 60:
```suggestion
```

tatiana (approved):
Great work, @pankajastro , happy for us to merge the PR once the docs (https://github.com/astronomer/dag-factory/blob/main/PRIVACY_NOTICE.md) are updated.

pankajastro:
> Great work, @pankajastro , happy for us to merge the PR once the docs (https://github.com/astronomer/dag-factory/blob/main/PRIVACY_NOTICE.md) are updated.

Updated the docs

Issue #272, opened 2024-10-22
[feature] Render YAML definitions in Airflow UI


tatiana:
**Context**

Even though the YAML is the source of truth for the DAG topology and the operators/tasks configuration, we cannot see those in the Airflow UI. This can make it harder to troubleshoot and confirm the behaviours are as expected or that the expected version was deployed.

**Acceptance criteria**

* [ ] Ideally, we'd display the whole YML in the Airflow UI (plugin?)
* [ ] If that's not possible, we should at least render as a templated field the YML representation for the task of interest

tatiana:
cc @cmarteepants

nightwyrm:
Thanks, @tatiana. We recently had dag-factory suggested to us, so I've been investigating it this weekend. Like you say, if we can't have those details surfaced, it makes it difficult to troubleshoot anything and that'll be a hard sell to our support engineers.

In a perfect world, the generated DAG code would be presented in the Code view (or custom tab), but even having the YAML accessible in some form would be great.

cmarteepants:
Thanks @rory-data, that makes sense for a workaround. What I'm hearing is that in a perfect world, you would see the serialized DAG in the code view versus the contents of the yaml file. Is that right?

rory-data:
@cmarteepants I would think so from a consistent or predictable experience perspective for support engineers. I do understand that it's not catered for in Airflow (Ash mentioned in Slack that it could be in AIP-85), so it may be a case of having the YAML presented in a better place than DAG Docs and using tagging to indicate a given DAG has been generated. That's just my two cents though 😅

rory-data:
I admittedly only came up with the idea in the weekend just gone as part of our initial investigations into using DAG Factory, so it's not being used in any live processes. 

However, I can see that a large YAML input could be potentially unwieldy in the DAG Docs field from a UX perspective with it being an expandable section high in the DAG UI. The way I've currently got it also means it would overwrite any user-supplied doc_md values, so I need to adjust it to append instead.

rory-data:
It needs polish, but FWIW, this is what I did for the [example_dag_factory.py](https://github.com/astronomer/dag-factory/blob/main/dev/dags/example_dag_factory.py) script in the examples directory. 

```python
"""
Airflow currently cannot serialise a DAG generated by DagFactory into the Code view of
the UI (this may be resolved in AIP-85).
As a workaround, the below change to the process of generating DAGs will render the YAML
[...]

PR #305, opened 2024-12-02
Render YML DAG config as DAG Docs

pankajastro:
Until DAG Factory 0.20, even though the YAML is the source of truth for the DAG topology and the operators/tasks configuration, the YAML was not displayed in the Airflow UI. This can make it harder to troubleshoot and confirm that the behaviours are as expected or that the expected version was deployed.

This PR solves this limitation by adding or appending the YAML file to the DAG `md_docs` field. The following image illustrates this feature being used via the Airflow UI:
Screenshot 2024-12-02 at 11 06 15 PM

Closes: https://github.com/astronomer/dag-factory/issues/272

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -801,10 +801,11 @@ class DagBuilder:
         # Render YML DAG in DAG Docs
         if self._yml_dag:
             subtitle = "## YML DAG"
+
             if dag.doc_md is None:
-                dag.doc_md = subtitle + "\n```yaml\n" + self._yml_dag
+                dag.doc_md = f"{subtitle}\n```yaml\n{self._yml_dag}\n```"
             else:
-                dag.doc_md = dag.doc_md + subtitle + "\n```yaml\n" + self._yml_dag
+                dag.doc_md += f"\n\n{subtitle}\n```yaml\n{self._yml_dag}\n```"
 
         # tags parameter introduced in Airflow 1.10.8
         if version.parse(AIRFLOW_VERSION) >= version.parse("1.10.8"):
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -805,7 +805,7 @@ class DagBuilder:
             if dag.doc_md is None:
                 dag.doc_md = f"{subtitle}\n```yaml\n{self._yml_dag}\n```"
             else:
-                dag.doc_md += f"\n\n{subtitle}\n```yaml\n{self._yml_dag}\n```"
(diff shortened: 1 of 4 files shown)


dagfactory/dagfactory.py
```diff
@@ -33,13 +33,33 @@ def __init__(self, config_filepath: Optional[str] = None, config: Optional[dict]
         self.dags_count: int = 0
         self.tasks_count: int = 0
         self.taskgroups_count: int = 0
+        self._config_filepath = config_filepath
```
pankajastro, line 36:
```suggestion
```

tatiana, line 36: Why do we need this?

dagfactory/dagfactory.py
```diff
@@ -33,13 +33,33 @@ def __init__(self, config_filepath: Optional[str] = None, config: Optional[dict]
...
             self.config: Dict[str, Any] = DagFactory._load_config(config_filepath=config_filepath)
         if config:
             self.config: Dict[str, Any] = config
 
+    @staticmethod
+    def _serialise_config_md(dag_name, dag_config, default_config):
+        # Remove empty task_groups if it exists
+        if dag_config.get("task_groups") == {}:
+            del dag_config["task_groups"]
+
```
tatiana, line 49: Is DAG factory injecting this empty task group on behalf of end-users?

dagfactory/dagbuilder.py
```diff
@@ -795,6 +798,15 @@ def build(self) -> Dict[str, Union[str, DAG]]:
...
             dag.doc_md = doc_md_callable(**dag_params.get("doc_md_python_arguments", {}))
 
+        # Render YML DAG in DAG Docs
+        if self._yml_dag:
+            subtitle = "## YML DAG"
+
+            if dag.doc_md is None:
+                dag.doc_md = f"{subtitle}\n```yaml\n{self._yml_dag}\n```"
+            else:
+                dag.doc_md += f"\n\n{subtitle}\n```yaml\n{self._yml_dag}\n```"
```
tatiana, line 808: Do we have a test covering this behaviour?

dagfactory/dagfactory.py
pankajastro: Added a comment: https://github.com/astronomer/dag-factory/pull/305/commits/08c52a2c2337d5dc8c23f3b169d8cecc4ddf49c9

tatiana (approved):
Thank you very much for working on this, @pankajastro , this will be very helpful for end-users. Once Airflow 3.0 is released and we know more about the new plugin strategy, we can review the approach.

Issue #225, opened 2024-10-01
Expand() usage does not pickup .output if task with expand comes before in yaml def/dict keys of task referenced in expand


StephenOTT:
If you have something like 

```
Tasks:
   Task1:
      ...
      Dependencies: [Task2]
      Expand: 
         Some_arg: Task2.output

   Task2:
      ...
```

Then when dagbuilder parses the tank/duct is does not properly change the expand arg to a referenced Xcom and it remains as a are

tatiana:
Thanks a lot for reporting this issue, @StephenOTT ! Would you be willing to contribute a fix? We should add an example DAG that leverages this so we can avoid regressions, once we fix this.

PR #307, opened 2024-12-04
Support building DAGs out of topologically unsorted YAML files

tatiana:
Any YAML files that declare upstream tasks after downstream tasks, regardless of using dynamic task mapping, would fail.

Example of DAG that would fail:
```
test_expand:
  default_args:
    owner: "custom_owner"
    start_date: 2 days
  description: "test expand"
  schedule_interval: "0 3 * * *"
  default_view: "graph"
  tasks:
    process:
      operator: airflow.operators.python_operator.PythonOperator
      python_callable_name: expand_task
      python_callable_file: $CONFIG_ROOT_DIR/expand_tasks.py
      partial:
        op_kwargs:
          test_id: "test"
      expand:
        op_args:
          request.output
      dependencies: [request]
    request:
      operator: airflow.operators.python.PythonOperator
      python_callable_name: example_task_mapping
      python_callable_file: $CONFIG_ROOT_DIR/expand_tasks.py
```

In this example, the upstream (parent) task "request" is defined after the downstream (child) task "process". Before this change, this DAG would fail.

I implemented a solution to solve the problem that uses Kahn's algorithm to sort the tasks topologically:
https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm

It has asymptotic complexity O(N + D), where N is the total number of tasks, and D is the total number of dependencies. This complexity seems acceptable.

An alternative to the current approach would be to create all the tasks without dependencies as a starting point and add the dependencies once all tasks were made - similar to what we did in https://github.com/astronomer/astronomer-cosmos. However, this approach would require a bigger refactor of the DAG factory and may have issues with dynamic task mapping.

Closes: #225

### DIFF
--- a/tests/fixtures/dag_factory_kubernetes_pod_operator.yml
+++ b/tests/fixtures/dag_factory_kubernetes_pod_operator.yml
@@ -53,7 +53,7 @@ example_dag:
       task_id: 'passing-task'
       get_logs: True
       in_cluster: False
-      dependencies: ['task_1']
+      dependencies: []
     task_2:
       operator: airflow.contrib.operators.kubernetes_pod_operator.KubernetesPodOperator
       namespace: 'default'
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -852,19 +852,23 @@ class DagBuilder:
         (https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm)
 
         The complexity is O(N + D) where N: total tasks and D: number of dependencies.
+
+        :returns: topologically sorted list containing tuples (task name, task config)
         """
         # Step 1: Build the downstream (adjacency) tasks list and the upstream dependencies (in-degree) count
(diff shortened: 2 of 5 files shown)


tests/fixtures/dag_factory_kubernetes_pod_operator.yml
```diff
@@ -53,7 +53,7 @@ example_dag:
       task_id: 'passing-task'
       get_logs: True
       in_cluster: False
-      dependencies: ['task_1']
+      dependencies: []
```
pankajastro, line 56: do we need this empty dependency []?

pankajastro (approved):

tests/fixtures/dag_factory_kubernetes_pod_operator.yml
tatiana: Yes, this example had a circular dependency and this was not a valid DAG 🙈

PR #320, opened 2024-12-16
[Feature] Refactor and add support for schedule conditions in DAG configuration:

ErickSeo:
### Description
This feature introduces a enhancement to DAG scheduling in Airflow, enabling support for dynamic schedules based on dataset conditions. By leveraging dataset filters and logical conditions, users can now create more flexible and precise scheduling rules tailored to their workflows.

**Key Features**:

- Condition-Based Scheduling: Allows defining schedules using logical conditions between datasets (e.g., ('dataset_1' & 'dataset_2') | 'dataset_3'), enabling workflows to trigger dynamically based on dataset availability.

- Dynamic Dataset Processing: Introduced the process_file_with_datasets function to evaluate and process dataset URIs from external files, supporting both simple and condition-based schedules.

- Improved Dataset Evaluation: Developed the evaluate_condition_with_datasets function to transform dataset URIs into valid variable names and evaluate logical conditions securely.

**Workflow Example**:
Given the following condition:
```yaml
example_custom_config_condition_dataset_consumer_dag:
  description: "Example DAG consumer custom config condition datasets"
  schedule:
    file: $CONFIG_ROOT_DIR/datasets/example_config_datasets.yml
    datasets:  "((dataset_custom_1 & dataset_custom_2) | dataset_custom_3)"
  tasks:
    task_1:
      bash_command: "echo 'consumer datasets'"
```
```yaml
example_without_custom_config_condition_dataset_consumer_dag:
  description: "Example DAG consumer custom config condition datasets"
  schedule:
    datasets: "((s3://bucket-cjmm/raw/dataset_custom_1 & s3://bucket-cjmm/raw/dataset_custom_2) | s3://bucket-cjmm/raw/dataset_custom_3)"
  tasks:
    task_1:
      bash_command: "echo 'consumer datasets'"
```

```yaml
example_without_custom_config_condition_dataset_consumer_dag:
  description: "Example DAG consumer custom config condition datasets"
  schedule:
    datasets: 
      !or  
        - !and  
          - "s3://bucket-cjmm/raw/dataset_custom_1"  
          - "s3://bucket-cjmm/raw/dataset_custom_2"
        - "s3://bucket-cjmm/raw/dataset_custom_3"
  tasks:
    task_1:
      bash_command: "echo 'consumer datasets'"
```

The system evaluates the datasets, ensuring valid references, and schedules the DAG dynamically when the condition resolves to True.

**Example Use Case**:
Consider a data pipeline that processes files only when multiple interdependent datasets are updated. With this feature, users can create dynamic DAG schedules that automatically adjust based on dataset availability and conditions, optimizing resource allocation and execution timing.

Images:
Captura de tela 2024-12-16 181059
Captura de tela 2024-12-16 181103
Captura de tela 2024-12-16 181131


dagfactory/dagbuilder.py
```diff
@@ -624,6 +610,104 @@ def replace_expand_values(task_conf: Dict, tasks_dict: Dict[str, BaseOperator]):
...
+        :type condition_string: Optional[str]
+
+        :returns: The result of the condition evaluation if `condition_string` is provided, otherwise a list of `Dataset` objects.
+        :rtype: Any
+        """
+        is_airflow_version_at_least_2_9 = version.parse(AIRFLOW_VERSION) >= version.parse("2.9.0")
+        if condition_string and is_airflow_version_at_least_2_9:
+            map_datasets = utils.get_datasets_map_uri_yaml_file(file, datasets_filter)
+            dataset_map = {alias_dataset: Dataset(uri) for alias_dataset, uri in map_datasets.items()}
+            return eval(condition_string, {}, dataset_map)
```
tatiana, line 658: Using eval can lead to security risks, debugging, and maintainability challenges.

The most critical drawback of eval is that it executes arbitrary code. DAG Factory users can execute arbitrary code using `BashOperator` on Airflow worker nodes at a task execution level. However, with the proposed change, we're allowing end-users to execute arbitrary code at a scheduler / DAG processor level, so the damage caused can have more significant implications.

What do you think about implementing a function [ast](https://docs.python.org/3/library/ast.html) to resolve this expression? It has very few node types (and, or, parenthesis and the actual literals), so it should not be complex.

tatiana (commented):
Hi @ErickSeo, this feature is super exciting. It significantly improves how DAG factory handles datasets. I'm sorry for the delay in reviewing it. I've left some comments in-line and am happy to discuss alternatives.

If we're able to reach a consensus on the implementation, we can aim to release this feature in DAG Factory 0.22,  planned to be released on 10 January 2025.

dagfactory/__init__.py
```diff
@@ -3,7 +3,4 @@
 from .dagfactory import DagFactory, load_yaml_dags
 
 __version__ = "0.21.0"
-__all__ = [
-    "DagFactory",
-    "load_yaml_dags",
-]
+__all__ = ["DagFactory", "load_yaml_dags"]
```
tatiana, line 6: I was surprised with the linting changes on areas of the code that weren't necessarily changed. How were these raised..? It may be worth to isolate them in a separate PR.

dagfactory/dagbuilder.py
```diff
@@ -624,6 +610,104 @@ def replace_expand_values(task_conf: Dict, tasks_dict: Dict[str, BaseOperator]):
...
+
+        :returns: The result of the logical condition evaluation with URIs replaced by valid variable names.
+        :rtype: Any
+        """
+        dataset_map = {}
+        for uri in datasets_filter:
+            valid_variable_name = utils.make_valid_variable_name(uri)
+            condition_string = condition_string.replace(uri, valid_variable_name)
+            dataset_map[valid_variable_name] = Dataset(uri)
+        evaluated_condition = eval(condition_string, {}, dataset_map)
```
tatiana, line 632: The same concern I shared in another part of the code applies here, in regards to using `eval`.

dagfactory/dagbuilder.py
```diff
@@ -624,6 +610,104 @@ def replace_expand_values(task_conf: Dict, tasks_dict: Dict[str, BaseOperator]):
...
+        :type condition_string: str
+        :param datasets_filter: A list of dataset URIs to be evaluated in the condition.
+        :type datasets_filter: List[str]
+
+        :returns: The result of the logical condition evaluation with URIs replaced by valid variable names.
+        :rtype: Any
+        """
+        dataset_map = {}
+        for uri in datasets_filter:
+            valid_variable_name = utils.make_valid_variable_name(uri)
```
tatiana, line 629: Why are we representing `Dataset` as variables in Python? This seems to introduce additional complexity without clear benefits. The previous approach of instantiating `Dataset` objects seemed clearer and simpler.

dagfactory/dagbuilder.py
```diff
@@ -624,6 +610,104 @@ def replace_expand_values(task_conf: Dict, tasks_dict: Dict[str, BaseOperator]):
...
+        if condition_string and is_airflow_version_at_least_2_9:
+            map_datasets = utils.get_datasets_map_uri_yaml_file(file, datasets_filter)
+            dataset_map = {alias_dataset: Dataset(uri) for alias_dataset, uri in map_datasets.items()}
+            return eval(condition_string, {}, dataset_map)
+        else:
+            datasets_uri = utils.get_datasets_uri_yaml_file(file, datasets_filter)
+            return [Dataset(uri) for uri in datasets_uri]
+
+    @staticmethod
+    def configure_schedule(dag_params: Dict[str, Any], dag_kwargs: Dict[str, Any]) -> None:
```
tatiana, line 664: It is very valuable the refactoring you did, of isolating this logic in a dedicated method, thank you.

dev/dags/datasets/example_dag_datasets.yml
```diff
@@ -52,3 +52,24 @@ example_custom_config_dataset_consumer_dag:
     task_1:
       operator: airflow.operators.bash_operator.BashOperator
       bash_command: "echo 'consumer datasets'"
+
+example_custom_config_condition_dataset_consumer_dag:
+  description: "Example DAG consumer custom config condition datasets"
+  schedule:
+    file: $CONFIG_ROOT_DIR/datasets/example_config_datasets.yml
+    datasets: ['dataset_custom_1', 'dataset_custom_2', 'dataset_custom_3']
+    conditions: "((dataset_custom_1 & dataset_custom_2) | dataset_custom_3)"
```
tatiana, line 61: In Airflow, users must use a single argument to define the **schedule**, regardles of how the schedule is set: `schedule` or `schedule_interval` (Airflow 2.3 and earlier versions). 

Currently, DAG Factory (0.21) already splits this configuration into two:
- `schedule_interval` for cron-based schedules and cron presets
- `schedule` for dataset scheduling

With this PR, we add a third option within `schedule`: `conditions`.

Why should the YAML mapping have so many keys to represent the same configuration? 

Some drawbacks if we continue in this direction:
1. **Usage complexity**: DAG Factory has a more complex syntax to define schedules than the original Airflow Python DAG
2. **Implementation complexity**: how do we handle scenarios where users attempt to define multiple of these arguments? (probably the safest would be to error if they try to use more than one)
3. **Lack of extensibility**: where should we add [`DatasetOrTimeSchedule`](https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/timetable.html#dataset-event-based-scheduling-with-time-based-scheduling) support? Should we create another key in DAG Factory YAMLs, increasing the complexity?

It would be great if we could find a more unified approach to represent `schedule`.

dev/dags/datasets/example_dag_datasets.yml
```diff
@@ -52,3 +52,24 @@ example_custom_config_dataset_consumer_dag:
...
+  tasks:
+    task_1:
+      operator: airflow.operators.bash_operator.BashOperator
+      bash_command: "echo 'consumer datasets'"
+
+example_without_custom_config_condition_dataset_consumer_dag:
+  description: "Example DAG consumer custom config condition datasets"
+  schedule:
+    datasets: ['s3://bucket-cjmm/raw/dataset_custom_1', 's3://bucket-cjmm/raw/dataset_custom_2', 's3://bucket-cjmm/raw/dataset_custom_3']
+    conditions: "((s3://bucket-cjmm/raw/dataset_custom_1 & s3://bucket-cjmm/raw/dataset_custom_2) | s3://bucket-cjmm/raw/dataset_custom_3)"
```
tatiana, line 71: An alternative would be to try to represent the expressions "natively" in YAML. We could use logical operators like `!or` and `!and` to custom specific logic:
```
schedule:
  !or
    - !and
      - "s3://bucket-cjmm/raw/dataset_custom_1"
      - "s3://bucket-"cjmm/raw/dataset_custom_2"
    - "s3://bucket-cjmm/raw/dataset_custom_3)"
```

dev/dags/datasets/example_dag_datasets.yml
ErickSeo: Great observation. Do you believe it would be feasible to implement the schedule field to accept multiple types? Utilize the schedule field to accept both a list of datasets and a string with datasets and conditions?

Example 1 List:
schedule:
    datasets: ['dataset_custom_1', 'dataset_custom_2', 'dataset_custom_3']

Example 2 String:
schedule:
    datasets: "((dataset_custom_1 & dataset_custom_2) | dataset_custom_3)"

Example 3 Yaml Expression:
schedule:  
  !or  
    - !and  
      - "s3://bucket-cjmm/raw/dataset_custom_1"  
      - "s3://bucket-cjmm/raw/dataset_custom_2"  
    - "s3://bucket-cjmm/raw/dataset_custom_3"  

On one hand, we simplify the process, but on the other, it might make the field less intuitive for users

Do you think this tradeoff is worth it?

dev/dags/datasets/example_dag_datasets.yml
ErickSeo: I modified the `_load_config function`. 

```python
@staticmethod
def _load_config(config_filepath: str) -> Dict[str, Any]:
    """
    Loads YAML config file to dictionary

    :returns: dict from YAML config file
    """
    # pylint: disable=consider-using-with
    try:

        def __join(loader: yaml.FullLoader, node: yaml.Node) -> str:
            seq = loader.construct_sequence(node)
            return "".join([str(i) for i in seq])
        
        def or_constructor(loader: yaml.FullLoader, node: yaml.Node) -> str:
            seq = loader.construct_sequence(node)
            return " | ".join([f"({str(i)})" for i in seq])

        def and_constructor(loader: yaml.FullLoader, node: yaml.Node) -> str:
            seq = loader.construct_sequence(node)
            return " & ".join([f"({str(i)})" for i in seq])

        yaml.add_constructor("!join", __join, yaml.FullLoader)
        yaml.add_constructor("!or", or_constructor, yaml.FullLoader)
        yaml.add_constructor("!and", and_constructor, yaml.FullLoader)

        with open(config_filepath, "r", encoding="utf-8") as fp:
            config_with_env = os.path.expandvars(fp.read())
            config: Dict[str, Any] = yaml.load(stream=config_with_env, Loader=yaml.FullLoader)
    except Exception as err:
        raise DagFactoryConfigException("Invalid DAG Factory config file") from err
    return config
```

Do you think it could cause any unexpected behavior? It worked when I tested it locally.

dagfactory/__init__.py
ErickSeo: I used `black dagfactory`

ErickSeo: If I don't commit the file without running black, it ends up failing in the CI.
https://github.com/astronomer/dag-factory/actions/runs/12590049231/job/35090935848?pr=320

dev/dags/datasets/example_dag_datasets.yml
ErickSeo: It definitely got a lot simpler. I’ll document about it in the README once everything’s good to go.

tatiana: We've recently made significant changes to the documentation:
https://github.com/astronomer/dag-factory/tree/main/docs

They are being published here:
https://astronomer.github.io/dag-factory/dev/

If you could create a separate PR for the documentation, it would be amazing!

I added a follow-up ticket, so we don't forget
https://github.com/astronomer/dag-factory/issues/361

dev/dags/datasets/example_dag_datasets.yml
tatiana: Thank you very much, @ErickSeo . Let's release like this - and if there are issues, people will report and we can address them.

I don't usually encourage unbound `Exception`, but since the code was like this, I believe we should stick to this for now.

dagfactory/__init__.py
tatiana: We also use black, via pre-commit, but it probably has some different configuration. For future contributions, you can learn more about it in the section "Pre-Commit and linting" in the docs:
https://astronomer.github.io/dag-factory/dev/contributing/howto/#run-integration-tests

tatiana (approved):
This is a very exciting feature, @ErickSeo ! Thank you very much for adding support to a more powerful way of scheduling DAGs in DAG Factory.

Thank you for addressing all the feedback. I'm sure there will be use-cases we did not think about - but we can always iterate over them.

Issue #253, opened 2024-10-17
Enable Enhanced Callback Support Across All Callback Types


jroach-astronomer:
Continuing the work done in #209, add enhanced callback support across all callback types, both at the DAG-level, as well as at the task-level.

jroach-astronomer:
@tatiana, @pankajastro, please assign this issue to me.

jroach-astronomer:
Clear line-of-sight forward on the work that is work to be done in order to apply this new pattern to all callback types.

PR #322, opened 2024-12-17
Add Robust Support for Callbacks at Task and TaskGroup Level

jroach-astronomer:
Pull request to address #253, which aims to provide enhanced support for callback types at the **DAG**, `default_args`, **Task** and **TaskGroup** level. There are now four different ways that callbacks can be defined via `dag-factory`. Below, you'll also see an example of each.

1. By passing a string that points to a callable.
2. By passing a file path and file name that points to a callable, similar to above.
3. By providing a string that points to a callable, but with some parameters being provided at runtime.
4. Using callbacks from providers (like Slack).

### Passing a string that points to a callable
``` yaml
...
  task_groups:
    task_group_1:
      default_args:
        on_success_callback: print_hello.print_hello_from_callback
      dependencies: [task_1, task_2]
...
```

### Passing a file path and file name that points to a callable
``` yaml
...
    task_2:
      bash_command: "echo 2"
      on_success_callback_name: print_hello_from_callback
      on_success_callback_file: $CONFIG_ROOT_DIR/print_hello.py
      dependencies: [start]
...
```

### By providing a string that points to a callable, but with some parameters being provided at runtime
``` yaml
...
  on_failure_callback:
    callback: customized.callbacks.custom_callbacks.output_message
    param1: param1
    param2: param2
...
```

### Using callbacks from providers
``` yaml
...
    on_failure_callback:
      callback: airflow.providers.slack.notifications.slack.send_slack_notification
      slack_conn_id: slack_conn_id
      text: |
        :red_circle: Task Failed.
        This task has failed and needs to be addressed.
        Please remediate this issue ASAP.
      channel: "#channel"
...
```

Unit-tests have been added and updated appropriately.


jroach-astronomer:
@pankajastro, do you mind reviewing this, open to feedback?

cc: @tatiana

dev/dags/example_callbacks.yml
```diff
@@ -15,14 +15,25 @@ example_callbacks:
     callback: customized.callbacks.custom_callbacks.output_message
     param1: param1
     param2: param2
+  task_groups:
+    task_group_1:
+      default_args:
+        on_success_callback: $CONFIG_ROOT_DIR.print_hello.print_hello_from_callback
```
pankajkoti, line 21: This specification looks a bit off. I specify `ENV CONFIG_ROOT_DIR=/usr/local/airflow/dags/` to define in my Dockerfile. When this would be expanded here, it would appear as `/usr/local/airflow/dags/print_hello.print_hello_from_callback` which is off.

can you please share how have you specified the value for CONFIG_ROOT_DIR?


When I am trying to test this branch and this particular example DAG, I see a parsing error
```
Broken DAG: [/usr/local/airflow/dags/example_callbacks.py] Traceback (most recent call last):
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1324, in _find_and_load_unlocked
ModuleNotFoundError: No module named '/usr/local/airflow/dags/'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/astro/.local/lib/python3.12/site-packages/dagfactory/dagfactory.py", line 163, in clean_dags
    dags: Dict[str, Any] = self.build_dags()
                           ^^^^^^^^^^^^^^^^^
  File "/home/astro/.local/lib/python3.12/site-packages/dagfactory/dagfactory.py", line 130, in build_dags
    raise DagFactoryException(f"Failed to generate dag {dag_name}. verify config is correct") from err
dagfactory.exceptions.DagFactoryException: Failed to generate dag example_callbacks. verify config is correct
```

When I comment out this line, the DAG parses fine.

pankajkoti: Also this task group does not have any tasks associated, no?  Can we add some tasks to this taskgroup to verify callbacks for this taskgroup are applied and working?

pankajkoti:
@jroach-astronomer please let me know once you'd like me to re-review/test the pull request. Could you please also update the PR description after the changes especially with the example for `on_success_callback: $CONFIG_ROOT_DIR.print_hello.print_hello_from_callback`

jroach-astronomer:
@pankajkoti, resolved MC's, updated PR docs.

dev/dags/example_callbacks.yml
```diff
@@ -11,18 +14,43 @@ example_callbacks:
...
-      python_callable_name: succeeding_task
+      # operator: airflow.operators.bash_operator.BashOperator
+      # bash_command: "echo 0"
+      operator: airflow.operators.empty.EmptyOperator
+      on_success_callback_name: print_hello_from_callback
+      on_success_callback_file: $CONFIG_ROOT_DIR/print_hello.py
+    task_1:
+      operator: airflow.operators.bash_operator.BashOperator
+      bash_command: "echo 1"
+      on_success_callback: customized.callbacks.custom_callbacks.output_message
```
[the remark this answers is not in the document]
pankajkoti: also what is the planned interface.

Do we intend to continue accepting direct callback functions e.g. for on_success_callback arg or they would mandatorily need to be provided as values as nested arg `callable`? Or we plan to support both approaches?

dagfactory/dagbuilder.py
```diff
@@ -165,78 +165,56 @@ def get_dag_params(self) -> Dict[str, Any]:
...
+                if utils.check_dict_key(dag_params, callback_type):
+                    dag_params[callback_type]: Callable = self.set_callback(
+                        parameters=dag_params, callback_type=callback_type
+                    )
 
-        if utils.check_dict_key(dag_params["default_args"], "on_success_callback_name") and utils.check_dict_key(
-            dag_params["default_args"], "on_success_callback_file"
-        ):
+                # Then, check at the DAG-level for a file path and name
+                if utils.check_dict_key(dag_params, f"{callback_type}_name") and utils.check_dict_key(
```
pankajkoti, line 192: I’m wondering if we should continue supporting the `{callback_type}_name` + `{callback_type}_file` approach for callbacks. Are there scenarios where simply specifying `{callback_type}` might be insufficient or lack necessary support?

Unifying the approach to a single, straightforward method seems more maintainable and less confusing for users. Additionally, what happens if a user specifies both `on_success_callback` and `on_success_callback_name + on_success_callback_file`? Currently, it appears that `on_success_callback_name + on_success_callback_file` will override the `on_success_callback`, which isn’t transparent to the end user and could cause confusion.

While this isn’t part of your changes and has been a pre-existing issue, it might be worth discussing whether we could drop support for the `name+file` approach altogether or at least an error when users try to provide callbacks using both approaches. What do you think?

pankajkoti (approved):
This looks fantastic! Thank you for all the refactoring and for addressing my earlier questions and comments. I’ve shared an in-line thought about potentially dropping support for the `file+name` approach of specifying callbacks. It would be great to revisit this idea later, and we can plan to work on it in the next release since it’s outside the scope of this PR.

dev/dags/example_callbacks.yml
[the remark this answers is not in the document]
pankajkoti: @jroach-astronomer and I discussed this over a call and it seems we plan to support both approaches. We will also highlight this in our documentation for callables.

Issue #233, opened 2024-10-09
Support dag_display_name property


aliellis:
Hi there, correct me if im wrong but dag-factory doesn't seem to support the [dag_display_name](https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/example_dags/example_display_name.html) property?

expected config/behaviour:
```yaml
example_dag:
  default_args:
    owner: "a-person"
    start_date: '2024-10-08'
  schedule_interval: None
  catchup: False
  description: 'Responsible for running an example DAG'
  dag_display_name: "Pretty Example DAG"
  ```

We then get a DAG called "Pretty Example DAG" not "example_dag"

Looks like it can be tagged on anywhere around [here](https://github.com/astronomer/dag-factory/blob/main/dagfactory/dagbuilder.py#L675-L681)? Happy to submit an MR if thats the case :)

tatiana:
Hi @aliellis , indeed, it seems to be the case. Please, if you could submit a PR, it would be apreciated!

PR #326, opened 2024-12-31
Propogate provided dag_display_name to built dag

pankajkoti:
This PR propagates the `dag_display_name` value set in the YAML dag config to the built dag object.
The support for associating `dag_display_name` was added in Airflow 2.9.0, hence, we only associate it
for the supported versions.

Preview of the built dag with dag_display_name being set.
Screenshot 2024-12-31 at 5 40 29 PM

closes: #233


pankajastro (approved):

tatiana (approved):
It looks great, thanks a lot, @pankajkoti !

Issue #319, opened 2024-12-16
Add migration docs


pankajastro:
Migrate Python DAG to YML

PR #327, opened 2025-01-02
Add docs comparing Python and YAML-based DAGs

tatiana:
Add documentation illustrating how two Airflow Python DAGs (one using traditional operators, another using TaskFlow API) are represented in Python (standard Airflow) and YAML (DAG Factory).

We use real-world data: one DAG uses Hacker News, and the other uses PyPI stats.

Before we merge this PR, I think we should merge #328 first and rebase this one. I isolated the tooling changes (mkdocs and others) in PR #328.

Close: #319


docs/comparison/taskflow_api.md
```diff
@@ -0,0 +1,105 @@
...
+
+```
+| package_name      |   last_day |   last_month |   last_week |
+|:------------------|-----------:|-------------:|------------:|
+| apache-airflow    |     852242 |     28194255 |     6253861 |
+| astronomer-cosmos |     442531 |     13354870 |     3127750 |
+| dag-factory       |      10078 |       354085 |       77752 |
+```
+
+The main logic is implemented as plain Python functions in [pypi_stats.py](../../dev/dags/pypi_stats.py):
```
pankajkoti, line 29: When building docs locally, it gives a warning for the link like below:
```
WARNING -  Doc file 'comparison/taskflow_api.md' contains a link '../../dev/dags/pypi_stats.py', but the target '../dev/dags/pypi_stats.py' is not found among documentation files.
```
I am wondering if we could use absolute URLs for these links to get away from those Warning messages?

tatiana: Yes, it makes sense; thanks for doing this in the other docs: #331!
Just fixed in 516c922077424d7eebb155f39edf0bb1f2d532d3

docs/comparison/traditional_operators.md
```diff
@@ -0,0 +1,95 @@
...
+to declare the workflow: while one uses Python, the other uses YAML.
+
+
+## Goal
+
+Let's say we'd like to create a workflow that performs the following:
+
+1. Retrieve the top ten stories from Hacker News using the [Hacker News API](https://github.com/HackerNews/API).
+2. Fetch the details for the two top stories using the Hacker News API.
+3. Summarize the selected stories Markdown, using Python.
```
pankajkoti, line 19:
```suggestion
3. Summarize the selected stories as Markdown, using Python.
```

docs/comparison/traditional_operators.md
```diff
@@ -0,0 +1,95 @@
...
+
+![alt text](../static/example_hackernews_dagfactory_graph.png "YAML DAG Graph visualization")
+
+### Airflow Code view
+
+From an Airflow UI perspective, the content displayed in the "Code" view is the main difference between the two implementations. While Airflow renders the original Python DAG, as expected, in the case of the YAML DAGs, Airflow displays the Python file that references the DAG Factory YAML files:
+
+```title="example_load_yaml_dags.py"
+--8<-- "dev/dags/example_load_yaml_dags.py"
+```
```
pankajkoti, line 82:
```suggestion
```

Looking at the built docs, this block seems redundant to me as the image in the below `Code View: Alternative DAG Factory YAML` shows the same snippet.

docs/comparison/taskflow_api.md
```diff
@@ -0,0 +1,105 @@
...
+![alt text](../static/example_pypi_stats_dagfactory_mapped_tasks.png "YAML DAG mapped tasks")
+
+
+### Airflow Code view
+
+From an Airflow UI perspective, the content displayed in the "Code" view is the main difference between the two implementations. While Airflow renders the original Python DAG, as expected, in the case of the YAML DAGs, Airflow displays the Python file that references the DAG Factory YAML files:
+
+```title="example_load_yaml_dags.py"
+--8<-- "dev/dags/example_load_yaml_dags.py"
+```
```
pankajkoti, line 92:
```suggestion
```
Looking at the built docs, this block seems redundant to me as the image in the below Code View: Alternative DAG Factory YAML shows the same snippet.

pankajkoti (approved):
LGTM with real world data analysis here :) 

A couple of minor suggestions inline

docs/comparison/taskflow_api.md
tatiana: The advantage of the non-screenshot reference is that people can copy and paste more quickly if they'd like to 🤔  Do you have any other suggestions to avoid the duplicate?

docs/comparison/traditional_operators.md
tatiana: Just replied in: https://github.com/astronomer/dag-factory/pull/327/files#r1901720084

Once we agree on the approach, I'll fix it in both places!

docs/comparison/taskflow_api.md
pankajkoti: No strong opinion there Tatiana! Happy to be it as it is now

PR #329, opened 2025-01-02
Add Markdown pre-commit checks

tatiana:
- Linting
- Codespell (spelling)
- Dead link check

And fixes problems identified by them when running:
```
pre-commit run --all-files
```


mkdocs.yml
pankajkoti, line 1: do we wish to remove this file?

pankajkoti (approved):
LGTM, thanks for adding these checks. Also, would like to seek attention on whether we need to restore the `mkdocs.yml` file that's currently getting removed with this PR.

mkdocs.yml
tatiana: Nice catch, @pankajkoti , it was removed by mistake - just restored it in d403a021cc450b4477f61479fceabf05e3fd694c. Thank you!

Issue #297, opened 2024-11-21
[Feature] Shared defaults for load_yaml_dags


wearpants:
### Description

It'd be nice to pass some shared `default_args` for a directory, either via a python object or a `defaults.yml` file in the directory.

### Use case/motivation

One DAG per file is easier for users IMO, and as a system administrator I'd like to be able to give them a shared set of pre-baked defaults (env vars, etc.)

### Related issues

#289, #290 

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

cmarteepants:
Something like this?

./dags/default.yml
``` 
default:
  catchup: false
  default_args:
    start_date: "2024-01-01"
  schedule_interval: "0 0 * * *"
  tasks:
    extract:
      operator: airflow.operators.python.PythonOperator
      python_callable_file: /usr/local/airflow/include/etl_helpers.py
      python_callable_name: extract_helper
    load:
      dependencies:
      - transform
      operator: airflow.operators.python.PythonOperator
      python_callable_file: /usr/local/airflow/include/etl_helpers.py
      python_callable_name: load_helper
    transform:
      dependencies:
      - extract
      op_kwargs:
        ds_nodash: '{{ds_nodash}}'
      operator: airflow.operators.python.PythonOperator
      python_callable_file: /usr/local/airflow/include/etl_helpers.py
      python_callable_name: transform_helper
```

./dags/bi.yml
``` 
business_analytics:
  schedule_interval: "@daily"
  tasks:
    load:
      op_kwargs:
        database_name: BA
        table_name: inventory
```

./dags/ds.yml
```
data_science:
  tasks:
    load:
      op_kwargs:
        database_name: DS
        table_name: daily_sales
```

./dags/ml.yml
```
machine_learning:
  tasks:
    load:
      op_kwargs:
        database_name: ML
        table_name: training_data
```
...

jroach-astronomer:
@cmarteepants, only thing I think I'd add here is referencing the default values in `.dags/bi.yml`, etc.

wearpants:
@cmarteepants So basically `bi.yml` etc are merged on top of `default.yml`? Could you clarify how that works - does that happen for the entire yaml object tree key-by-key / lists extended / etc? How would you do overrrides? (Take a look at [ChainMap](https://docs.python.org/3/library/collections.html#collections.ChainMap) for a simple comparison).

[...]

PR #330, opened 2025-01-02
Support loading default_args from shared defaults.yml

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/297

Currently, DAG Factory provides two places to configure default_args:

1. At the top of the YML file.
2. In the DAG configuration YML.

The second option overrides the first one.

Post this PR, the user can also keep the default_args in the `defaults.yml` file. The configuration from  `defaults.yml`  will be applied to all DAG Factory DAG

Sample `defaults.yml`
```
default_args:
  start_date: "2025-01-01"
  owner: "global_owner"
  depends_on_past: true
```

The precedence for default_args will be as follows, after this implementation:

1. At DAG configuration YML
2. At the top of the YML file
3. default.yml

i.e At DAG configuration YML will take precedence.

### DIFF
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -45,19 +45,14 @@ class DagFactory:
             with open(default_args_yml, "r") as file:
                 return yaml.safe_load(file)
 
-    def deep_merge(self, dict1, dict2):
-        for key, value in dict2.items():
-            if isinstance(value, dict) and key in dict1 and isinstance(dict1[key], dict):
-                dict1[key] = self.deep_merge(dict1[key], value)
-            else:
-                dict1[key] = value
-        return dict1
-
     def _merge_default_args(self):
         global_default_args = self._global_default_args()
         default_config: Dict[str, Any] = self.get_default_config()
 
-        return self.deep_merge(global_default_args, default_config)
+        merged_config = global_default_args.copy()
+        merged_config.update(default_config)
+
+        return merged_config
 
     @staticmethod
     def _serialise_config_md(dag_name, dag_config, default_config):
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -13,6 +13,8 @@ from airflow.models import DAG
 from dagfactory.dagbuilder import DagBuilder
 from dagfactory.exceptions import DagFactoryConfigException, DagFactoryException
 
+# rom dagfactory.utils import merge_configs
+
 # these are params that cannot be a dag name
 SYSTEM_PARAMS: List[str] = ["default", "task_groups"]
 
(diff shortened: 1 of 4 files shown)


dagfactory/dagfactory.py
```diff
@@ -36,6 +37,25 @@ def __init__(self, config_filepath: Optional[str] = None, config: Optional[dict]
         if config:
             self.config: Dict[str, Any] = config
 
+    @staticmethod
+    def _global_default_args():
+        default_args_yml = Path(settings.DAGS_FOLDER) / "defaults.yml"
```
tatiana, line 42: There are cases when`defaults.yml` would not be placed in the root of the DAGs folder.

Currently, DAG Factory has two main entry points:
1. Use `dagfactory.load_yaml_dags` with a directory. DAG Factory will inspect the given directory recursively. It defaults to the Airflow `DAGS_FOLDER` using `airflow_conf.get("core", "dags_folder")` - but it may not be the same.
2. Use `dagfactory.DagFactory` with a single YAML file. In this case, DAG Factory would be inspecting that specific YAML file.

To be consistent with the existing behaviour of DAG Factory, perhaps we should retrieve the default args from `load_yaml_dags`, if given, otherwise we default to Airflow's `DAGS_FOLDER`, WDYT?

It is probably worth to change https://github.com/astronomer/dag-factory/blob/177f60859b140e22d339307d6316665eb93657db/dagfactory/dagfactory.py#L181 to use the same way of retrieving the user-defined DAGs folder.

dagfactory/dagfactory.py
```diff
@@ -36,6 +37,25 @@ def __init__(self, config_filepath: Optional[str] = None, config: Optional[dict]
...
+                return yaml.safe_load(file)
+
+    def _merge_default_args(self, global_default_args, local_default_args):
+        result = global_default_args.copy()
+
+        for key, value in local_default_args.items():
+            if isinstance(value, dict) and key in result and isinstance(result[key], dict):
+                result[key] = self._merge_default_args(result[key], value)
+            else:
+                result[key] = value
```
tatiana, line 55: Did you have any particular use-case in mind to add this logic?

I feel we should not attempt to merge the values of the "duplicate" keys in different default configurations. This could lead to more confusion.

We can override keys at the top level, non-recursively, following our order of precedence.

dagfactory/dagfactory.py
```diff
@@ -13,6 +13,8 @@
 from dagfactory.dagbuilder import DagBuilder
 from dagfactory.exceptions import DagFactoryConfigException, DagFactoryException
 
+# rom dagfactory.utils import merge_configs
+
```
pankajastro, line 17:
```suggestion
```

dagfactory/dagfactory.py
pankajastro: I thought we have nested default_args, but we don't have nested default_args. I have removed this method and just updating now

dagfactory/dagfactory.py
pankajastro: I have exposed `default_args_config_path` in method `load_yaml_dags` with default `airflow_conf.get("core", "dags_folder") `

tatiana (approved):
Looks good, thanks a lot, @pankajastro!

PR #336, opened 2025-01-03
Doc updates

yanmastin-astro:


tatiana (approved):
Thank you for the super fast review, @yanmastin-astro, and improvements!

Issue #279, opened 2024-10-30
Host user-facing DAG Factory docs


tatiana:
To be aligned with @cmarteepants and the docs team where we want to host the DAG Factory docs:
1. ReadTheDocs
2. Astro website
3. GitHub pages

pankajkoti:
Since this is also related to #278 , and @pankajastro you might be aligning with @cmarteepants & the docs team on this one too, I am assigning this issue to you too. Please feel to re-assign/un-assign  it later when we would actually take care of the implementation for this based on our Sprint planning.

cmarteepants:
This is timely! Astro docs are moving away from netlify, so doesn't make sense to host on Astro until that's done. Whatever we use has to be markdown-compatible for an easy lift to the new framework, which excludes Sphinx/ReadTheDocs since that uses .rst. 

Their recommendation is to use Github Pages (and markdown!)

pankajastro:
We have decided to host the initial version of docs on GitHub Pages

PR #342, opened 2025-01-06
Add GH action job to deploy docs

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/279

This PR includes the following changes

1. Add CI job to deploy docs
2. Introduced command `hatch run docs:dev` to locally build and serve docs
3. Remove `hatch run docs:serve` command
4. Rename `hatch run docs:build` to `hatch run docs: gh_deploy`
5. Rename `hatch run docs:release` to `hatch run docs: gh_release`


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,44 @@
...
+          python-version: "3.12"
+          architecture: "x64"
+
+      - name: Install packages and dependencies
+        run: pip install hatch
+
+      - name: Deploy Docs
+        run: |
+          git config user.name "Pankaj Singh"
+          git config user.email "pankaj.singh@astronomer.io"
```
pankajastro, line 36: @tatiana / @pankajkoti Do we have any bot for this that we can use

This look like may need: https://github.com/jimporter/mike?tab=readme-ov-file#deploying-via-ci

pyproject.toml
```diff
@@ -114,9 +114,9 @@ dependencies = [
 ]
 
 [tool.hatch.envs.docs.scripts]
-build = "mike deploy --push dev"
-serve = "mike serve"
-release = "version=$(python -W ignore -c 'import dagfactory; print(dagfactory.__version__)') && mike deploy --push --update-aliases $version latest"
+dev = "mkdocs build && mkdocs serve" # For local development and preventing publishing
```
pankajastro, line 117: Changing this to avoid publishing from local machine

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,44 @@
+name: CI Docs
+
+on:
+  push: # Run on pushes to the default branch
+    branches: [main, docs_job]
```
pankajastro, line 5:
```suggestion
    branches: [main]
```

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,44 @@
...
+  pull_request_target: # Also run on pull requests originated from forks
+    branches: [main, docs_job]
```
pankajastro, line 7:
```suggestion
    branches: [main]
```

pyproject.toml
```diff
@@ -114,9 +114,9 @@ dependencies = [
...
+gh_deploy = "mike deploy --push dev"
+gh_release = "version=$(python -W ignore -c 'import dagfactory; print(dagfactory.__version__)') && mike deploy --push --update-aliases $version latest"
```
pankajkoti, line 119:
```suggestion
gh-deploy = "mike deploy --push dev"
gh-release = "version=$(python -W ignore -c 'import dagfactory; print(dagfactory.__version__)') && mike deploy --push --update-aliases $version latest"
```
Can we use `-` separators. I guess they suit well for commands similar to how we have for `test-cov`, `test-integration`, `static-check`, etc defined earlier in this file.

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Deploy Docs
+        run: |
+          git config user.name "Pankaj Singh"
+          git config user.email "pankaj.singh@astronomer.io"
+
+          git fetch origin gh-pages --depth=1
+
+          if [[ $GITHUB_EVENT_NAME == "release" && $GITHUB_EVENT_ACTION == "published" ]]; then
+              hatch run docs:gh_release
```
pankajkoti, line 41:
```suggestion
              hatch run docs:gh-release
```

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,44 @@
...
+          else
+              hatch run docs:gh_deploy
```
pankajkoti, line 43:
```suggestion
              hatch run docs:gh-deploy
```

pankajkoti (approved):
It looks great to me @pankajastro . Have some minor suggestions in-line.

pyproject.toml
tatiana: Thanks a lot, @pankajastro !

.github/workflows/docs.yaml
tatiana: I believe it is very important we use a bot for this - if we don't have one yet, we can create one. Please, confirm with @phanikumv , @pankajastro - we should make this change before merging.

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Install packages and dependencies
+        run: pip install hatch
+
+      - name: Deploy Docs
+        run: |
+          git config user.name "Pankaj Singh"
+          git config user.email "pankaj.singh@astronomer.io"
+
+          git fetch origin gh-pages --depth=1
```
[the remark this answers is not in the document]
pankajkoti: would it make sense to have plain fetch instead of only fetching the latest commit? i mean removing --depth=1 option?

.github/workflows/docs.yaml
phanikumv: yes, we shouldnt use personal details. @pankajastro could you request IT to create a `oss-integrations-bot` user. Then we can add it to the repo and test it

pankajkoti:
> * When we merge PRs to the main branch, I believe we should update the "dev" or "unstable" docs only
> * We should probably update the "versioned" documents only when we publish a new version of DAG Factory to PyPI

I believe these 2 are already taken care of, no?

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,46 @@
...
+          python-version: "3.12"
+          architecture: "x64"
+
+      - name: Install packages and dependencies
+        run: pip install hatch
+
+      - name: Deploy Docs
+        run: |
+          # git config user.name "Pankaj Singh"
+          # git config user.email "pankaj.singh@astronomer.io"
```
tatiana, line 38: @pankajastro would it be an alternative to set the following Github Action environment variable?
```
        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Use GitHub token for authentication
```

.github/workflows/docs.yaml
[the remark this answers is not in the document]
pankajastro: This would be a little bit fast, and docs also suggest that entire history does not require: https://github.com/jimporter/mike?tab=readme-ov-file#deploying-via-ci

.github/workflows/docs.yaml
pankajastro: I have changed it to dummy user and look like working

.github/workflows/docs.yaml
pankajastro: Look like it also work with dummy user :)

.github/workflows/docs.yaml
tatiana: Would it be worth to check if this would work as well?

pankajastro: Testing looks like user.name and email require

tatiana (approved):
Thanks a lot for addressing all the feedback, @pankajastro , I'm super excited to see the improvements we're making to the docs published - with versions!

tatiana:
@pankajastro please, before merging this PR, update the docs: https://github.com/astronomer/dag-factory/blob/main/docs/contributing/howto.md#write-docs

PR #347, opened 2025-01-09
Enable Depandabot to scan outdated Github Actions dependencies

tatiana:
Recently, a high manager at Astronomer received a message saying "Subject: [Action required] Actions artifacts v3 is closing down", with a list of repos within the org that had outdated dependencies.

To avoid this in the future, we're enabling Github's Depandabot, which will automatically create PRs suggesting these fixes proactively - so this situation does not happen again.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


pankajkoti (approved):
Amazing, this is great! Was unaware of this 🚀

Issue #379, opened 2025-03-10
[Feature]  Support for inlets in dag-factory


IvanSviridov:
### Description

Currently, dag-factory only supports defining outlets, but there is no built-in support for inlets. This limitation prevents proper integration with lineage tracking services such as DataHub, OpenLineage, and other metadata-driven tools.
Why is this needed?
- Airflow 2.4.0+ introduced Dataset-based scheduling, where both inlets and outlets play a critical role in tracking data dependencies.
- Many data lineage tools (e.g., DataHub, Marquez, OpenLineage) rely on **both inlets** and outlets to correctly infer dataset relationships.
- Without inlets, lineage propagation in Airflow **does not fully capture** data source dependencies, which results in incomplete lineage graphs.

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!
Related Pull Request: #380

pankajkoti:
@IvanSviridov looking forward to your contribution! :)

PR #380, opened 2025-03-10
Add support for defining inlets

IvanSviridov:
DAG Factory 0.22  only supports defining outlets, but there is no built-in support for inlets. This limitation prevents proper integration with lineage tracking services such as DataHub, OpenLineage, and other metadata-driven tools.

Why is this needed?
- Airflow 2.4.0+ introduced Dataset-based scheduling, where both inlets and outlets play a critical role in tracking data dependencies.
- Many data lineage tools (e.g., DataHub, Marquez, OpenLineage) rely on **both inlets** and outlets to correctly infer dataset relationships.
- Without inlets, lineage propagation in Airflow **does not fully capture** data source dependencies, which results in incomplete lineage graphs.

Closes:  #379


tatiana (commented):
This is looking great, @IvanSviridov , is there any chance you could add some tests and an example YAML? If you could extend our docs with this feature, it would also be amazing.

IvanSviridov:
@tatiana, could you check it again please
and it is issue https://github.com/astronomer/dag-factory/issues/379

tatiana (approved):
This looks great, thanks for adding tests and updating the docs, alongside the feature itself, @IvanSviridov !

if possible, please, could you just remove the `dags` symbolic link? This feels like an unrelated change.

dags
```diff
@@ -0,0 +1 @@
+dev/dags
```
tatiana, line 1: It is probably worth removing this symbolic link, since it seems unrelated to the proposed feature.

tatiana: I just deleted the symbolic link, looks good now

matveykortsev:
@tatiana Hi! When you are planning to release this? Is it possible to release as minor release of 22x?

pankajastro:
Hi @matveykortsev, We have released a pre-release version, [dag-factory 0.23.0a1](https://pypi.org/project/dag-factory/0.23.0a1/), which is expected to include this PR.

Issue #370, opened 2025-01-22
[Bug] Issue with HttpOperator Data Handling in DAGFactory


a-chumagin:
### DAG Factory version

0.22.0

### airflow version

2.10.2

### Python version

3.12

### Deployment

Docker-Compose

### What happened?

I am experiencing an issue when using dag-factory to configure HttpOperator tasks. The request body is incorrectly serialized, and the content type is being treated as form data instead of JSON. This issue occurs only when the task is created via DAGFactory, but works fine when using raw HttpOperator code.

### Relevant log output

```shell

```

### How to reproduce

1. Create a DAG using DAGFactory with an HttpOperator task that sends JSON data.
2. Ensure the data is correctly formatted as a JSON string (e.g., data: '{"data_source": "postgres"}').
3. When the task runs, check the request body sent to the HTTP endpoint.
4. The body is sent as form-encoded data (e.g., data_source=postgres), not as JSON.

**Expected Behavior:**
The request body should be serialized as valid JSON and sent with the Content-Type: application/json header.

**Actual Behavior:**
The body is sent as form data (e.g., data_source=postgres), not as JSON. This causes the request to fail.

Example of YAML
```
basic_example_dag:
  default_args:
    owner: "@a.chumagin"
  description: "this is an example dag"
  schedule_interval: "0 3 * * *"
  tags: ['dqa']
  render_template_as_native_obj: True
  tasks:
    send_request:
      operator: airflow.providers.http.operators.http.HttpOperator
      http_conn_id: "dqp_host_dev"
      method: "POST"
      endpoint: "/run_test"
      data: '{"data_source": "postgres"}'
      headers:
        Content-Type: application/json
      log_response: True
```

The headers send correct . I use echo service for handle request 
```
POST /run_test HTTP/1.1
Host: echo-local.com:8087
Accept: */*
Accept-Encoding: gzip, deflate
Content-Length: 20
Content-Type: application/json
User-Agent: Python/3.12 aiohttp/3.10.5
data_source=postgres
```

I tried different ways to present data 
1. As dictionary 
```
   data: 
          data_source: "postgres"
```
2. With """ and > wrappers. Same result

When I use raw HttpOperator without DagFactory the issues does not occur 
```
HttpOperator(task_id='http_soda_check',
                                   http_conn_id='dqp_host_dev',
                                    method='POST',
                                    endpoint="/run_test",
                                    data="""
                                    {
                                    "data_source": "postgres"
                                   }
                                   """,
[...]

PR #382, opened 2025-03-25
Add HttpOperator JSON serialization support with tests

a-chumagin:
This PR resolved issue https://github.com/astronomer/dag-factory/issues/370

Add HttpOperator JSON serialization support with tests

- Add utility function for JSON serialization of HTTP request data
- Add automatic Content-Type header handling for HttpOperator
- Support both dictionary and string JSON inputs with validation  
- Add comprehensive tests for HttpOperator serialization
- Add YAML examples for HTTP operator configuration

Additionally: 
- Fix Makefine 

Tested by:
1. Running Airflow with DagFactory
2. Executing a DAG with HttpOperator tasks 
3. Verifying request body format with an echo-server
4. Confirming proper Content-Type headers are set


a-chumagin:
@tatiana hello! 
It looks like dependences issues between airflow and http libs. I try to fix this. 
Also for some reason could of failed tests locally and here are different .

dev/dags/example_simple_http_operator_task.yml
```diff
@@ -0,0 +1,35 @@
...
+  render_template_as_native_obj: True
+  tasks:
+    send_request_json:
+      operator: airflow.providers.http.operators.http.SimpleHttpOperator
+      http_conn_id: "example_host"
+      method: "POST"
+      endpoint: "/run_test"
+      data:
+        data: "fake_data"
+        format: "json"
```
pankajastro, line 21: do we need this format key here, how are we using this in the implementation?

a-chumagin:
## HTTP Operator Compatibility

### Airflow Version Challenges
Supporting HTTP operators across multiple Airflow versions required handling differences in operator naming and import paths:
- **Airflow 2.0 - providers-http  4.10.0**: `SimpleHttpOperator` in `airflow.providers.http.operators.http`
- **Airflow providers-http >= 4.11.0**: `HttpOperator` (renamed)

To ensure compatibility, we implemented dynamic imports:
```python
try:
    from airflow.providers.http.operators.http import HttpOperator
    HTTP_OPERATOR_CLASS = HttpOperator
except ImportError:
    try:
        from airflow.providers.http.operators.http import SimpleHttpOperator
        HTTP_OPERATOR_CLASS = SimpleHttpOperator
    except ImportError:
        HTTP_OPERATOR_CLASS = None
```

### Generalized Testing
To handle both `SimpleHttpOperator` and `HttpOperator`:
1. Parametrized tests were created to work with either operator.
2. Conditional test skipping was implemented for unsupported versions.
3. Dynamic fixture selection was used based on the operator class:
   ```python
   if HTTP_OPERATOR_PATH == "airflow.providers.http.operators.http.HttpOperator":
       fixture_path = "tests/fixtures/dag_factory_http_operator_task.yml"
   else:
       fixture_path = "tests/fixtures/dag_factory_simple_http_operator_task.yml"
   ```

### Integration Test Limitations
Integration testing faced the following constraints:
1. Complex CI setup was required to test across multiple Airflow versions.
2. Mock servers or specialized environments were needed to verify HTTP requests.
3. Serialization differences across versions required version-specific assertions.

Unit tests were prioritized to ensure compatibility while minimizing the need for complex integration setups.

dev/dags/example_simple_http_operator_task.yml
a-chumagin: It is a just example and I choose not clear key for json. 
It can be like 
```
name: "fake_name"
lastname: "fake_lastname"
```
I used key format:json for defining plain text and json format in tests

pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 ]
 dependencies = [
     "apache-airflow>=2.0",
-    "apache-airflow-providers-http",
+    "apache-airflow-providers-http>=2.0.0",
```
pankajastro, line 33: Any reason why we are upgrading this?

pankajastro (approved):

pyproject.toml
a-chumagin: As I understand from my investigation versions less then 2 does not make sense for this

pankajastro: @a-chumagin After giving it some more thought, I believe that CI should not fail even without this upgrade. Simply bumping the version without corresponding code cleanup seems like an incomplete change.

We’ve opened an issue https://github.com/astronomer/dag-factory/issues/387 to address the version bump of the HTTP provider.

Additionally, we observed some inconsistencies with the Airflow version in this project. We are running tests in CI for Airflow >= 2.2, while the dependency requires Airflow >= 2.0. To resolve this, we’ve created an issue https://github.com/astronomer/dag-factory/issues/385.

I kindly request that you revert this upgrade in this PR and submit a new one specifically for the version bump.

a-chumagin: @pankajastro I agree. PR should reflect only one problem :) I reverted version restriction . Let's wait tests

PR #394, opened 2025-04-08
Add script to check version and tag

pankajastro:
Add a script to check the version and tag in pypi publish job


tatiana (approved):
Looks like a safe approach, @pankajastro !
Did any previous issue motivate this change?

Issue #381, opened 2025-03-20
[Feature] DagFactory.__init__ should support default_args_config dict


wearpants:
### Description

Similar to passing a dict for `config` instead of a path in `config_filepath`, it should be possible to pass a `default_args_config` dict, not only a filepath in `default_args_config_path`

https://github.com/astronomer/dag-factory/blob/main/dagfactory/dagfactory.py#L31

### Use case/motivation

I generate python dicts for `config` from a yaml file with some custom preprocessing. I'd like to be able to do the same dict creation for `defaults.yml`, but there is no way to pass a dict for this purpose, only a filepath.

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

wearpants:
Relevant touchpoints (it's a bit of spaghetti code):

1.https://github.com/astronomer/dag-factory/blob/main/dagfactory/dagfactory.py#L45
1. https://github.com/astronomer/dag-factory/blob/61beb2908be67dabf941224188c77be641d8cef2/dagfactory/dagfactory.py#L133
1. https://github.com/astronomer/dag-factory/blob/61beb2908be67dabf941224188c77be641d8cef2/dagfactory/dagfactory.py#L136

jroach-astronomer:
I'd be happy to take a closer look at this one.

cc: @pankajastro

wearpants:
Any thoughts? I have a bit of budget $$$ I could allocate to help move development forward

jroach-astronomer:
@wearpants, are you looking to pass this `dict` in at the DAG-level, or the project-level?

jroach-astronomer:
@wearpants, a PR has been created to address this: https://github.com/astronomer/dag-factory/pull/404

wearpants:
@jroach-astronomer this seems to do what I wanted (passing default_args per DAG). Thank you, much appreciated!

PR #404, opened 2025-05-01
Providing option to set global default_args via a dictionary

ghost:
This PR aims to address #381, which will allow for a user to pass in a dictionary of `default_args` when  instantiating their DAGs. The alternative is to leverage a specific YAML file (we'll fall back to `defaults.yml`).

How is this different than our current functionality? Rather than relying solely on a function, users can define `default_args` when they generate their DAGs in a `.py` file.


jroach-astronomer:
@pankajastro, if you get a chance, please review this PR!

pankajastro (commented):
Could you please an example and document this feature?

dagfactory/dagfactory.py
```diff
@@ -26,24 +26,50 @@ class DagFactory:
     :type config_filepath: str
     :param config: DAG factory config dictionary. Cannot be user with `config_filepath`.
     :type config: dict
+    :param default_args_config_path:
```
pankajastro, line 29: Could you please add the missing details about this param?

dagfactory/dagfactory.py
```diff
@@ -26,24 +26,50 @@ class DagFactory:
...
+    :type default_args_config_path: str
+    :param default_args:
```
pankajastro, line 31: same here

jroach-astronomer: Resolved.

dagfactory/dagfactory.py
jroach-astronomer: Resolved.

jroachgolf84:
@pankajastro, please feel free to close this PR, I'll go ahead and open a new one with my updated user.

dagfactory/dagfactory.py
```diff
@@ -24,26 +24,52 @@ class DagFactory:
...
+    :param default_args: A dictionary of default arguments for that DAG, as an alternative to default_args_config_path.
+    :type default_args: dict
     """
 
     def __init__(
         self,
         config_filepath: Optional[str] = None,
         config: Optional[dict] = None,
         default_args_config_path: str = airflow_conf.get("core", "dags_folder"),
+        default_args: Optional[dict] = None,  # By default, this is None (we'll fall back to default_args_config_path)
```
pankajastro: shall we rename this to default_args_config_dict?

jroachgolf84: Renamed!

Issue #288, opened 2024-11-18
[Bug] Import error using latest Airflow Kubernetes provider package


Orpheuz:
### DAG Factory version

0.20.0

### airflow version

2.8.1

### Python version

3.11

### Deployment

Docker-Compose

### What happened?

The newly released version of [apache-airflow-providers-cncf-kubernetes](https://pypi.org/project/apache-airflow-providers-cncf-kubernetes/#history) includes the deprecation of `airflow.providers.cncf.kubernetes.operators.kubernetes_pod` which ultimately throws an exception in `dagbuilder.py` imports:

### Relevant log output

```shell
Broken DAG: [/usr/local/airflow/dags/dag_generator.py] Traceback (most recent call last):
  File "/usr/local/airflow/.local/lib/python3.11/site-packages/dagfactory/dagbuilder.py", line 66, in <module>
    from airflow.providers.cncf.kubernetes.operators.kubernetes_pod import KubernetesPodOperator
ModuleNotFoundError: No module named 'airflow.providers.cncf.kubernetes.operators.kubernetes_pod'
```

### How to reproduce

Run dag-factory (0.20.0) with the latest kubernetes Airflow provider

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

pankajastro:
Could you please set apache-airflow-providers-cncf-kubernetes<10.0.0 as a dependency in your project?  kubernetes_pod has been removed in apache-airflow-providers-cncf-kubernetes version 10.0.0 see changelog https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/changelog.html#id1

Orpheuz:
Yes, I added that as a workaround

pankajastro:
Hey @Orpheuz, could you please confirm if you're still encountering this error? Also, could you share your DAG Python/YML example so we can try to reproduce it?

pankajastro:
I believe the PR: https://github.com/astronomer/dag-factory/pull/311 has fixed the issue.

PR #407, opened 2025-05-26
Upgrade KPO

pankajastro:


jlaneve (commented):
I wonder if there's a way we could make the Kubernetes provider an optional dependency - that'd be nice for folks who don't want to use it

pankajastro:
> I wonder if there's a way we could make the Kubernetes provider an optional dependency - that'd be nice for folks who don't want to use it

Good point! I believe this should be possible. I've created an issue to track it: https://github.com/astronomer/dag-factory/issues/409

dagfactory/dagbuilder.py
```diff
@@ -267,10 +266,21 @@ def make_timetable(timetable: str, timetable_params: Dict[str, Any]) -> Timetabl
             raise DagFactoryException(f"Failed to import timetable {timetable} due to: {err}") from err
         try:
             schedule: Timetable = timetable_obj(**timetable_params)
-        except Exception as err:
+        except Exception as err:  # pragma: no cover
```
tatiana, line 269: Would it be possible for us to capture a more specific exception here, instead of `Exception`? I understand it was like this before, but it may be an opportunity to improve this.

dagfactory/dagbuilder.py
```diff
@@ -369,77 +379,103 @@ def make_task(operator: str, task_params: Dict[str, Any]) -> BaseOperator:
...
+
+                if version.parse(K8S_PROVIDER_VERSION) >= version.parse("7.8.0"):
+                    # See PR: https://github.com/apache/airflow/pull/35063
+                    task_params["host_aliases"] = (
+                        [V1HostAlias(**v) for v in task_params.get("host_aliases")]
+                        if task_params.get("host_aliases") is not None
                         else None
                     )
 
+                task_params["tolerations"] = (
```
tatiana, line 441: This method is quite complex, and we repeatedly apply the same logic for each parameter.

Would this be an opportunity to extract the logic for the K8S pod operator in a separate function and try to reuse a bit of the code that seems duplicated? Do we need to set as `None` values that are not defined in the YAML?

Some thoughts on a potentially refactored version (not tested):

```
key_to_class = {
    "tolerations": V1Toleration,
    "security_context": V1PodSecurityContext,
    ...
}

for param_key, param_values in task_params.items():
    task_params[param_key] = key_to_class[param_key](param_values)
```

tatiana (commented):
This looks great, @pankajastro , it's impressive how many tickets we're closing with this change.

With this PR, are there any versions of Airflow and the K8s provider that we stop supporting?  It would be great to highlight this in the PR description and release notes.

Rob-murphys:
I am trying this pre-release but now get issues similar to #288 even though I am using `apache-airflow-providers-cncf-kubernetes==10.4.1`

```
Traceback (most recent call last):
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1324, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'airflow.providers.cncf.kubernetes.operators.kubernetes_pod'
```

dagfactory/dagbuilder.py
pankajastro: I have tried to clean it a bit in commit: 07387cabf267cc786da49da0ba40792071801591

dagfactory/dagbuilder.py
pankajastro: I noticed this pattern appears in a few places throughout the repo. Would it be okay if I addressed it in a separate PR? I've created an issue to track it here: https://github.com/astronomer/dag-factory/issues/410.

pankajastro:
> With this PR, are there any versions of Airflow and the K8s provider that we stop supporting? It would be great to highlight this in the PR description and release notes.

This should for version >= 4.x

tatiana (approved):
Excellent work, @pankajastro , thank you very much for addressing all the feedback! I'm happy for the further refactor to happen in a follow up PR.

Issue #399, opened 2025-04-15
Fix `schedule` and `schedule_interval` logic for airflow3 support


pankajastro:
The `schedule` and `schedule_interval` settings depend on these kwargs. Since `schedule_interval` has been removed in Airflow 3, we need to update or refactor the code below (and any related code) accordingly.

https://github.com/astronomer/dag-factory/blob/7ca60067aec7037b132f03308a902383c3a56bb5/dagfactory/dagbuilder.py#L718-L759

PR #413, opened 2025-05-30
Fix schedule args for Airflow 3

pankajastro:
Closes: https://github.com/astronomer/dag-factory/issues/399
Documentation PR: https://github.com/astronomer/dag-factory/pull/424

This PR adds support for handling the schedule parameter across various types of scheduling formats in DAG configuration. The changes in this PR is only apply to Airflow 3 users.

For detailed usage examples and documentation, please refer to:
https://github.com/astronomer/dag-factory/pull/424

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -788,7 +788,7 @@ class DagBuilder:
         if schedule is None:
             return None
 
-        if schedule.strip().lower() == "none":
+        if isinstance(schedule, str) and schedule.strip().lower() == "none":
             return None
 
         # Case 1: If schedule is a string, return it directly


dagfactory/dagbuilder.py
```diff
@@ -723,6 +723,45 @@ def process_file_with_datasets(file: str, datasets_conditions: str) -> Any:
             datasets_uri = utils.get_datasets_uri_yaml_file(file, list(dataset_map.keys()))
             return [Dataset(uri) for uri in datasets_uri]
 
+    @staticmethod
+    def _get_schedule_obj(dag_params: Dict[str, Any]):
+        schedule = dag_params.get("schedule")
```
jlaneve, line 787: I wonder if we should also get `schedule_interval` if `schedule` doesn't exist. Could be nice to use this as a compatibility layer

pankajastro: Yes, we'll take care of that. This functionality is specific to AF3. If AF2 is installed in the environment, the previous logic will continue to work as expected.

tests/test_dagbuilder.py
```diff
@@ -1158,3 +1160,282 @@ def test_multiple_dependencies(self):
         assert task_names.index("task1") < task_names.index("task3")
         assert task_names.index("task2") < task_names.index("task4")
         assert task_names.index("task3") < task_names.index("task4")
+
+    @pytest.mark.skipif(INSTALLED_AIRFLOW_VERSION.major < 3, reason="Require at least Airflow 3.0.0")
+    def test_asset_schedule(self):
```
tatiana: @pankajastro would it be worth to break the tests into smaller ones, so it is easier to list all possible failures and/or isolate errors?

dagfactory/dagbuilder.py
```diff
@@ -726,6 +726,85 @@ def process_file_with_datasets(file: str, datasets_conditions: str) -> Any:
...
+            if op == "or":
+                return reduce(lambda a, b: a | b, assets)
+            elif op == "and":
+                return reduce(lambda a, b: a & b, assets)
+            else:
+                raise ValueError(f"Unknown operator: {op}")
+
+        if isinstance(value, dict):
+            if "or" in value:
+                return _combine_assets([DagBuilder._asset_schedule(item) for item in value["or"]], op="or")
```
tatiana: Python has [known performance issues](https://www.designgurus.io/answers/detail/why-is-python-recursion-so-expensive-and-what-can-we-do-about-it) with recursion. Would there be an alternative approach to implement this?

dagfactory/dagbuilder.py
```diff
@@ -726,6 +726,85 @@ def process_file_with_datasets(file: str, datasets_conditions: str) -> Any:
             datasets_uri = utils.get_datasets_uri_yaml_file(file, list(dataset_map.keys()))
             return [Dataset(uri) for uri in datasets_uri]
 
+    @staticmethod
+    def _asset_schedule(value):
```
tatiana: This function is quite complex. Would it be possible to break it down into smaller functions, that could be individually tested and changed having a smaller impact?

dagfactory/dagbuilder.py
```diff
@@ -726,6 +726,85 @@ def process_file_with_datasets(file: str, datasets_conditions: str) -> Any:
...
+                return timedelta(**value)
+
+            elif schedule_type == "relativedelta":
+                from dateutil.relativedelta import relativedelta
+
+                return relativedelta(**value)
+
+            elif schedule_type == "assets":
+                return DagBuilder._asset_schedule(value)
+            else:
```
tatiana: So far, the DAG Factory implementation has heavily relied on lots of ifs/else blocks. Could we improve this somehow, e.g. by using a dictionary with the key and specific function, or perhaps by isolating in a class, or any other strategy that may simplify the code and make it easier to maintain?

tatiana (commented):
It's really exciting to see all the new scheduling support in DAG Factory, @pankajastro , while still being backwards compatible. Excellent work! I left some comments inline.

dagfactory/dagbuilder.py
pankajastro: I have tried to simplify it in commit: 30eadfc1f457e1ac0fa7c96d778580a7782023ea

dagfactory/dagbuilder.py
pankajastro: I've kept the current recursive approach, as we discussed on the call. I believe the recursion depth will remain small in this case, so performance impact should be minimal. I've also created [issue #438](https://github.com/astronomer/dag-factory/issues/438) to track performance data for future evaluation.

tests/test_dagbuilder.py
pankajastro: I have tried to split this in commit: 16bb041a112e2fdaf5bb589f2063ecc808c8e534 PTAL

dagfactory/dagbuilder.py
pankajastro: I have tried to split this in commit: https://github.com/astronomer/dag-factory/commit/16bb041a112e2fdaf5bb589f2063ecc808c8e534 PTAL

tests/test_dagbuilder.py
pankajastro: Pushed a fix in commit: df35bfae1a9e86ae9ac22cf78807e471c702ef18

tatiana (approved):
It looks great, thanks a lot, @pankajastro !

PR #420, opened 2025-06-10
Add support for new single-file YAML DAG format with list-based tasks

jlaneve:
I've added support for a new single-file format, where the file is named `<dag-id>.dag.yaml`. This can be used like so:

```python
import os
from pathlib import Path

from dagfactory import load_single_dag, load_single_dag_directory

DEFAULT_CONFIG_ROOT_DIR = "/usr/local/airflow/dags/"

# Example 1: Load a single DAG file
dag_file = str(CONFIG_ROOT_DIR / "example.dag.yaml")
load_single_dag(globals_dict=globals(), dag_file_path=dag_file)

# Example 2: Load all DAGs from a directory
dags_dir = str(CONFIG_ROOT_DIR / "single_dags")
load_single_dag_directory(globals_dict=globals(), dags_folder=dags_dir)
```

Then, your yaml file (`example.dag.yaml`) is:

```yaml
# The DAG ID is taken from the filename (example.dag.yaml)

# DAG parameters
default_args:
  owner: airflow
  start_date: 2025-01-01
  end_date: 2025-01-01

schedule_interval: "@daily"
description: "Example DAG using the new single-file format"

tasks:
  - task_id: task_1
    bash_command: "echo 1"
  - task_id: task_2
    bash_command: "echo 2"
    dependencies: [task_1]
```

I've also changed the tasks part of the DAG to be either a dict (current behavior) _or_ a list (new behavior).

### Future Work

We should use the [Airflow magic loop](https://medium.com/apache-airflow/airflows-magic-loop-ec424b05b629) to only load the dag that's needed, which we can identify by file name.


tatiana:
@jlaneve I strongly believe DAG Factory should adopt a single, consistent YAML standard. Supporting multiple YAML interfaces introduces unnecessary maintenance complexity and documentation overhead.

The proposal in this PR is more restrictive than the current interface, and for that reason, I'm inclined to vote against moving forward with its implementation.

Specifically, the proposed single-file format does not support multiple workflows per file—a capability that has been available in DAG Factory since v0.22.0 and earlier.

Moreover, as Airflow continues to evolve, it is embracing more flexible authoring patterns, including the ability to define standalone assets without requiring a DAG/task structure. I believe we should focus on expanding the existing YAML interface to support these emerging use cases, allowing users to define assets and/or DAGs in the way that best fits their needs.

PR #424, opened 2025-06-17
Add scheduling docs for Airflow 3 

pankajastro:
While working on [PR #413](https://github.com/astronomer/dag-factory/pull/413), I noticed that the schedule parameter is currently inconsistent and lacks clear documentation. It's difficult to determine what formats are supported without diving into the codebase.

This PR proposes changes to make the schedule configuration more predictable, consistent, and easier to understand for users.

### DIFF
--- a/docs/configuration/schedule.md
+++ b/docs/configuration/schedule.md
@@ -1,4 +1,4 @@
-# Scheduling in DAG-Factory
+# Scheduling
 
 - *Released in version: 0.23.0*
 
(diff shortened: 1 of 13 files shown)


docs/configuration/schedule.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Example Overview
+
+| Type            | Description                           | Use Case Example                    |
+| --------------- | ------------------------------------- |-------------------------------------|
+| `cron`          | Run based on a cron string            | Every day at midnight               |
+| `timedelta`     | Fixed intervals between runs          | Every 6 hours                       |
+| `relativedelta` | Calendar-aware schedule (e.g. months) | Every 1st of the month              |
+| `timetable`     | Advanced Airflow timetables           | Custom trigger logic                |
+| `assets`        | Trigger based on asset readiness      | When data `X` and `Y` are available |
```
pankajastro, line 23: Add dataset

docs/configuration/schedule.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Schema Options
+
+### 1. Cron-Based Schedule
+
+```yaml
+schedule:
+  type: cron
+  value: "0 0 * * *"
+```
```
cmarteepants: For CRON, could we have a shorthand to omit the type? i.e

```
schedule: "0 0 * * *"
```

docs/configuration/schedule.md
```diff
@@ -0,0 +1,117 @@
...
+  type: assets
+  value:
+    and:
+      - uri: s3://dag1/output_1.txt
+        extra:
+          hi: bye
+      - uri: s3://dag2/output_1.txt
+        extra:
+          hi: bye
+```
```
cmarteepants: URI's are no longer mandatory in 3.0. We also have the "name" arg. 

Similar to CRON, wondering if we can somehow simplify this to avoid having to put a type for simple cases. If all I want is to schedule against `s3://dag1/output_1.txt` it's a lot of boilerplate to put the type. Lesser of a priority than for CRON though.

docs/configuration/schedule.md
pankajastro: Makes sense! This one would be easy to handle

docs/configuration/schedule.md
pankajastro: They can also pass the name parameter. I'll update the example. I'll see if we can simplify it for some basic use cases, as you are suggesting.

docs/configuration/schedule.md
```diff
@@ -0,0 +1,130 @@
...
+      extra:
+        hi: bye
+      watchers:
+        - class: airflow.sdk.AssetWatcher
+          name: test_asset_watcher
+          trigger:
+            class: airflow.providers.standard.triggers.file.FileDeleteTrigger
+            params:
+              filepath: "/temp/file.txt"
+```
```
tatiana, line 86: Would it be worth to reference a code that's running in the CI, instead of hard-coding it here? The CI would give us confidence that the example is working as expected, if we're able to add some cross-reference.

tatiana (approved):
Looks great, @pankajastro , minor feedback inline.

docs/configuration/schedule.md
```diff
@@ -0,0 +1,130 @@
+# Scheduling in DAG-Factory
+
+DAG-Factory offers flexible scheduling options so your DAGs can run on time or based on data availability. Whether you're triggering DAGs on a cron schedule, a time delta, or when an upstream asset is ready, you can configure it easily using the schedule field in your YAML.
```
tatiana, line 5: Would it be worth to state from which version tihs is working?
This may not be needed, since we now have versioned docs.

pankajastro: Fixed in: d0931ac21a12fbc27ba36f91d22dc562d68785fa

docs/configuration/schedule.md
pankajastro: Fixed in commit: 200b6e3ff87f7311467f576372e2a9e8c509c8f4

PR #432, opened 2025-06-24
Add setting CONFIG_ROOT_DIR in the contribution doc

gyli:
This PR fixes the contribution doc by completing the necessary env variables if Airflow standalone mode is used.

Although `DEFAULT_CONFIG_ROOT_DIR` is also used in some example config files, it is not required, hence not added in this PR.


pankajastro (approved):
Thanks, @gyli, for improving this!

Issue #426, opened 2025-06-19
Support for user-defined complex property types in YAML (e.g., !!k8s.V1ResourceRequirements)


pankajastro:
Enable end-users to explicitly define the "type" of complex properties in YAML configuration files using Python class annotations. This will help in scenarios where users want to override advanced objects (e.g., Kubernetes client models like k8s.V1ResourceRequirements) that need to be declared in configuration files and parsed into specific Python types.

### Example Use Case

```yaml
container_resources: !!k8s.V1ResourceRequirements
  - requests
    - cpu: "1000m"
    - memory: "8G"
  - limits
    - cpu: "16000m"
    - memory: "128G"
```
This should be parsed and instantiated as a k8s.V1ResourceRequirements object with the corresponding fields populated.

Possible Solution. 

Add a mechanism for YAML deserialization to recognise and instantiate tagged Python objects using the `!!module.ClassName` syntax.

Issue #417, opened 2025-06-05
[Bug] dag factory broken on KubernetesExecutor


RNHTTR:
### DAG Factory version

0.22.0

### airflow version

2.10.5 and 2.11.0

### Python version

default astro runtime python version

### Deployment

Astronomer

### What happened?

Worker pods fail without spinning up

### Relevant log output

```shell

```

### How to reproduce

try running dag factory on astro runtime 12.5 or 13.0 with k8s exec

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Contact Details

you know where to find me ;)

Issue #376, opened 2025-02-23
[Bug] executor_config not working when using Kubernetes Executor


nicolasge:
### DAG Factory version

0.22

### airflow version

2.10.5

### Python version

3.12

### Deployment

Official Apache Airflow Helm Chart

### Deployment details

Nothing special, the deployment is running on an openshift cluster using Kubernetes Executor only. 

### What happened?

1. All tasks with executor_config are not able to run, scheduler complains: INVALID executor_config config
2. When we use executor_config, we just want to override some of the specs(for instance resource limits), but dag-factory tried to clean all the executor_config generated by Scheduler, then use the one we defined.  It should be inherit then override, now it's clean flush then set only the values we defined.

### Relevant log output

```shell

```

### How to reproduce

Any task with executor_config pod_override settings, for instance resource limits.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

pankajastro:
Hi @nicolasge, could you please give the latest pre-release dag-factory==0.23.0a5 a try? If the issue persists, could you also share your executor_config?

PR #433, opened 2025-06-24
Add support for custom type yaml

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/426
closes: https://github.com/astronomer/dag-factory/issues/376
closes: https://github.com/astronomer/dag-factory/issues/417
closes: https://github.com/astronomer/dag-factory/issues/104

Support custom type object in yaml DAG. 

TODO: 
- Add tests
- Add docs
- Add support for more object

Example DAG

```
example_custom_yaml_type:
  description: "this is an example dag"
  schedule_interval: "@daily"
  render_template_as_native_obj: True
  tasks:
    task_1:
      bash_command: "echo 1"
      executor_config:
        pod_override: !!kubernetes.client.models.V1Pod
          spec: !!kubernetes.client.models.V1PodSpec
            containers:
                - !!kubernetes.client.models.V1Container
                  name: "base"
                  resources: !!kubernetes.client.models.V1ResourceRequirements
                    limits:
                        "cpu": "1"
                        "memory": "1024Mi"
                    requests:
                        "cpu": "0.5"
                        "memory": "512Mi"
    task_2:
      bash_command: "echo 2"
      dependencies: [task_1]
    task_3:
      bash_command: "echo 2"
      dependencies: [task_1]
```

Screenshot 2025-06-24 at 12 35 58 PM

### DIFF
--- a/dagfactory/__init__.py
+++ b/dagfactory/__init__.py
@@ -2,7 +2,7 @@
 
 from .dagfactory import DagFactory, load_yaml_dags
 
-__version__ = "0.23.0a4"
+__version__ = "0.23.0a6"
 __all__ = [
     "DagFactory",
     "load_yaml_dags",
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -12,6 +12,7 @@ from airflow.models import DAG
 
 from dagfactory.dagbuilder import DagBuilder
 from dagfactory.exceptions import DagFactoryConfigException, DagFactoryException
+from dagfactory.loader import DAGFactoryLoader
 
 # these are params that cannot be a dag name
 SYSTEM_PARAMS: List[str] = ["default", "task_groups"]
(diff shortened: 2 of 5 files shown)


tatiana (commented):
Hi @pankajastro , leaving some early feedback, as you asked for it.

I really like the idea of leveraging YAML tags, and how extensible it is, to allow end-users deciding on how to convert resources. When you add the docs, please, include an example of how an end-users could add a different resource type that is not defined in DAG Factory and how they could extend DAG Factory interface to know how to construct that element given that type.

My only current concern is: I believe the YAML should be easily parsable for any YAML parsing tools.

I validated the [current YAML](https://github.com/astronomer/dag-factory/blob/146c9902b9051f2b1f47ac2a5a1062f5ba98490d/dev/dags/example_custom_yaml_type.yml) using a few tools, and they all failed:
* https://yamlchecker.com/
* https://www.yamllint.com/
* https://jsonformatter.org/yaml-validator
* https://onlineyamltools.com/validate-yaml

If we can come up with a strategy where those tools can not fail when they face a Tag unknown for them, it would be great. Most end-users will have YAML linters and/or pre-commit hook checks which we want to continue working after this change.

pankajastro:
> Hi @pankajastro , leaving some early feedback, as you asked for it.
> 
> I really like the idea of leveraging YAML tags, and how extensible it is, to allow end-users deciding on how to convert resources. When you add the docs, please, include an example of how an end-users could add a different resource type that is not defined in DAG Factory and how they could extend DAG Factory interface to know how to construct that element given that type.
> 
> My only current concern is: I believe the YAML should be easily parsable for any YAML parsing tools.
> 
> I validated the [current YAML](https://github.com/astronomer/dag-factory/blob/146c9902b9051f2b1f47ac2a5a1062f5ba98490d/dev/dags/example_custom_yaml_type.yml) using a few tools, and they all failed:
> 
> * https://yamlchecker.com/
> * https://www.yamllint.com/
> * https://jsonformatter.org/yaml-validator
> * https://onlineyamltools.com/validate-yaml
> 
> If we can come up with a strategy where those tools can not fail when they face a Tag unknown for them, it would be great. Most end-users will have YAML linters and/or pre-commit hook checks which we want to continue working after this change.


Thank you very much for the feedback, @tatiana. I truly appreciate it.

These checkers weren’t part of my initial validation process, but I did test the DAG yaml using the custom parser implemented in this PR. After your feedback, I validated the YAML files with the mentioned checkers and, as you pointed out, they are indeed failing — except for [yamllint.com](https://www.yamllint.com/), which passed.

I believe this behavior is expected, as these tools are likely designed to avoid executing custom code for security reasons. Since our implementation uses custom types during parsing, allowing execution could introduce potential security risks, especially in general-purpose validation tools.

That said, the YAML is accepted by several other validators, including:
- https://goteleport.com/resources/tools/yaml-validator/
- https://www.yamllint.com/
- https://codebeautify.org/yaml-validator

I also referred to the YAML specification, which recommends the **failsafe schema** for broad compatibility. It’s the only schema guaranteed to be processed by any YAML parser, and therefore recommended for use in generic YAML tools.

> The failsafe schema is guaranteed to work with any YAML [document](https://yaml.org/spec/1.2.2/#documents). It is therefore the recommended [schema](https://yaml.org/spec/1.2.2/#recommended-schemas) for generic YAML tools. A YAML [processor](https://yaml.org/spec/1.2.2/#processes-and-models) should therefore support this [schema](https://yaml.org/spec/1.2.2/#recommended-schemas), at least as an option.

https://yaml.org/spec/1.2.2/#failsafe-schema

Our syntax aligns with the spec as well — particularly in regard to [local tags](https://yaml.org/spec/1.2.2/#tags).

If the concern is around pre-commit hooks or linting, it’s worth noting that most of these tools should ideally validate syntax only, rather than serializing or deserializing custom objects. For instance, [this pre-commit YAML checker](https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/check_yaml.py) is already enabled in our CI and appears to be functioning correctly.

Additionally, I manually tested using `yaml.load_all(...)` and `yaml.parse(...)`, which are the methods used by the plugin. These worked as expected.

I understand your concern but, I also believe it's important to highlight the potential security risks of allowing any external tool to execute unknown custom constructors during parsing.

pankajastro:
closing in favour of https://github.com/astronomer/dag-factory/pull/444

PR #436, opened 2025-06-25
Enable Airflow 3 tests in cI

pankajastro:


dagfactory/dagbuilder.py
```diff
@@ -771,7 +774,11 @@ def configure_schedule(dag_params: Dict[str, Any], dag_kwargs: Dict[str, Any]) -
                 if has_datasets_attr:
                     schedule.pop("datasets")
         else:
-            dag_kwargs["schedule"] = dag_params.get("schedule")
+            schedule = dag_params.get("schedule")
+            if schedule == "None":
+                dag_kwargs["schedule"] = None
+            else:
+                dag_kwargs["schedule"] = schedule
```
pankajastro: The schedule will fix in PR: https://github.com/astronomer/dag-factory/pull/413

.github/workflows/cicd.yaml
```diff
@@ -3,7 +3,7 @@ name: CI jobs
 on:
   # run on pushes to the default branch
   push:
-    branches: [main]
+    branches: [main, af3_ci]
```
```suggestion
    branches: [main]
```

dagfactory/dagbuilder.py
```diff
@@ -771,7 +774,11 @@ def configure_schedule(dag_params: Dict[str, Any], dag_kwargs: Dict[str, Any]) -
                 if has_datasets_attr:
                     schedule.pop("datasets")
         else:
-            dag_kwargs["schedule"] = dag_params.get("schedule")
+            schedule = dag_params.get("schedule")
+            if schedule == "None":
```
pankajkoti:
```suggestion
            if schedule.strip().lower() == "none":
```

pankajastro: Fixed in 4ce9d03fea05a8d876de06c2fd1d2f5dcaaa25cf

tests/utils.py
```diff
@@ -182,3 +188,46 @@ def _get_or_create_dagrun(
 
 def one_hour_ago(execution_date: datetime):
     return execution_date - datetime.timedelta(hours=1)
+
+
+def get_http_sensor():
```
tatiana:
```suggestion
def get_http_sensor_path():
```

tests/utils.py
```diff
@@ -182,3 +188,46 @@ def _get_or_create_dagrun(
...
+
+def get_schedule_key():
+    airflow_version = version.parse(AIRFLOW_VERSION)
+    if airflow_version < version.parse("3.0.0"):
+        return "schedule_interval"
+    else:
+        return "schedule"
+
+
+def get_bash_operator():
```
tatiana:
```suggestion
def get_bash_operator_path():
```

tests/utils.py
```diff
@@ -182,3 +188,46 @@ def _get_or_create_dagrun(
...
+    airflow_version = version.parse(AIRFLOW_VERSION)
+    if airflow_version >= version.parse("3.0.0"):
+        return "airflow.providers.standard.operators.bash.BashOperator"
+    elif airflow_version >= version.parse("2.4.0"):
+        return "airflow.operators.bash.BashOperator"
+    else:
+        return "airflow.operators.bash_operator.BashOperator"
+
+
+def get_python_operator():
```
tatiana:
```suggestion
def get_python_operator_path():
```

tests/utils.py
```diff
@@ -182,3 +188,46 @@ def _get_or_create_dagrun(
...
+    airflow_version = version.parse(AIRFLOW_VERSION)
+    if airflow_version >= version.parse("3.0.0"):
+        return "airflow.providers.standard.operators.python.PythonOperator"
+    elif airflow_version >= version.parse("2.4.0"):
+        return "airflow.operators.python.PythonOperator"
+    else:
+        return "airflow.operators.python_operator.PythonOperator"
+
+
+def get_sql_sensor():
```
tatiana:
```suggestion
def get_sql_sensor_path():
```

dagfactory/utils.py
```diff
@@ -345,3 +345,41 @@ def get_json_serialized_callable(data_obj):
         raise TypeError(f"data_obj must be a dict or str, not {type(data_obj)}")
 
     return lambda **kwargs: serialized_json
+
+
+def update_yaml_structure(data):
```
tatiana, line 348: @pankajastro I believe this method is extremely valuable, and we should "advertise" it in the documentation as part of the Airflow 3 migration. I'm happy for the docs to be done in a follow-up ticket/PR, but if we could log it, it would be great.

Also, it would be great if we could split our example YAML dags into two directories:
- airflow2
- airflow3

And have the DAGs materialised so that, without having they can be converted in real time.

Ideally, we'd educate DAG Factory users on using the new YAML standard, rather than having to convert it for them.

dagfactory/dagfactory.py
```diff
@@ -108,6 +115,12 @@ def __and(loader: yaml.FullLoader, node: yaml.Node) -> str:
             with open(config_filepath, "r", encoding="utf-8") as fp:
                 config_with_env = os.path.expandvars(fp.read())
                 config: Dict[str, Any] = yaml.load(stream=config_with_env, Loader=yaml.FullLoader)
+
+                # This will only invoke in the CI
+                # Make yaml DAG compatible for Airflow 3
+                if version.parse(AIRFLOW_VERSION) >= version.parse("3.0.0") and os.getenv("TEST_MODE"):
```
tatiana: @pankajastro, what if we had a configuration "AUTO_CONVERT_TO_PY3" or something like this, instead of `TEST_MODE`, and make it part of the public DAG Factory interface?

I'm usually a bit sceptical about it since it means the CI is running things differently, and we may be hiding other issues.

It would be great if we ran tests for both "AUTO_CONVERT_TO_PY3" values, True and False.

pankajastro: I'm checking Airflow version to test actual yaml files for Airflow version other than 3

The think I'm changing at runtime very predictable 
- schedule_interval to -> schedule
-   And update operator path if operator belongs to key of dict
```
operator_map = {
        "airflow.operators.dummy_operator.DummyOperator": "airflow.providers.standard.operators.empty.EmptyOperator",
        "airflow.operators.bash.BashOperator": "airflow.providers.standard.operators.bash.BashOperator",
        "airflow.operators.bash_operator.BashOperator": "airflow.providers.standard.operators.bash.BashOperator",
        "airflow.operators.python_operator.PythonOperator": "airflow.providers.standard.operators.python.PythonOperator",
        "airflow.operators.python.PythonOperator": "airflow.providers.standard.operators.python.PythonOperator",
    }
```
Happy to change to env name

dagfactory/utils.py
pankajastro: This tool i think require some more work if we want to expose to end user. I have created issue: https://github.com/astronomer/dag-factory/issues/439 to fix. 

> Also, it would be great if we could split our example YAML dags into two directories:
>
> - airflow2
> - airflow3

Yes, I agree but the PR was already large and also I wanted to keep the minimum almost duplicated yaml file. But we can improve on it.

dagfactory/dagfactory.py
pankajastro: I have changed env to AUTO_CONVERT_TO_AF3

dagfactory/utils.py
tatiana: I'm happy for this to be done on a follow-up ticket/PR, thanks, @pankajastro!

tatiana (approved):

PR #440, opened 2025-06-26
Remove expandvars in utils.get_python_callable

gyli:
`expandvars` was initially added to `get_python_callable` to avoid the hard coded path for Py files https://github.com/astronomer/dag-factory/pull/116. Then env var support is added to the entire YAML content https://github.com/astronomer/dag-factory/pull/236. This makes `expandvars` no longer needed in the `get_python_callable` function, hence this PR removes it for better performance.


tatiana (approved):
Nice catch, @gyli , thanks for your contribution!

PR #449, opened 2025-06-30
Add DAG example showcasing runtime params usage

pankajastro:
Add DAG example showcasing runtime params usage

Relevant discussion:
- https://apache-airflow.slack.com/archives/C03T0AVNA6A/p1750268447189009
- https://astronomer.slack.com/archives/C07CNMS6TKL/p1745588483040239

Screenshot 2025-06-30 at 12 25 58 PM

Screenshot 2025-06-30 at 12 26 25 PM

### DIFF
--- a/dev/dags/sample.py
+++ b/dev/dags/sample.py
@@ -1,5 +1,6 @@
 from datetime import datetime, timedelta
 from random import randint
+from typing import Any
 
 try:
     from airflow.providers.standard.operators.python import get_current_context
@@ -53,5 +54,7 @@ def one_day_ago(execution_date: datetime):
     return execution_date - timedelta(days=1)
 
 
-def read_params(params):
+def read_params(params: dict[str, Any]) -> None:
     print("params: ", params)
+    print("model_version:", params["model_version"])
+    print("my_param:", params["my_param"])
--- a/dev/dags/sample.py
+++ b/dev/dags/sample.py
@@ -51,3 +51,7 @@ def extract_last_name(full_name: str):
 
 def one_day_ago(execution_date: datetime):
     return execution_date - timedelta(days=1)
+
+
+def read_params(params):
+    print("params: ", params)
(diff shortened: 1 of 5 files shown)


dev/dags/sample.py
```diff
@@ -51,3 +51,7 @@ def extract_last_name(full_name: str):
 
 def one_day_ago(execution_date: datetime):
     return execution_date - timedelta(days=1)
+
+
+def read_params(params):
+    print("params: ", params)
```
pankajkoti, line 58: Would be nice to show how to access one of the params here

docs/configuration/params.md
```diff
@@ -0,0 +1,20 @@
+# Params
+
+**Params** in Airflow are a way to pass dynamic runtime configuration to tasks within a DAG. They allow tasks to be more flexible by enabling templated values to be injected during execution.
```
pankajkoti: Maybe we could add a link to the Airflow docs for Params feature

dev/dags/sample.py
```diff
@@ -51,3 +51,7 @@ def extract_last_name(full_name: str):
 
 def one_day_ago(execution_date: datetime):
     return execution_date - timedelta(days=1)
+
+
+def read_params(params):
```
pankajkoti: If possible could we add a type annotation for this arg params?

dev/dags/sample.py
pankajastro: good idea, added in commit: e2290ef03e7ea925d7d6c2ae1a735312dc0ce1db

docs/configuration/params.md
pankajastro: Added link in commit: e2290ef03e7ea925d7d6c2ae1a735312dc0ce1db

dev/dags/sample.py
pankajastro: Added type in: e2290ef03e7ea925d7d6c2ae1a735312dc0ce1db

PR #452, opened 2025-07-01
Support env var in default

gyli:
This PR enables YAML constructor, type casting, and env var support to `defaults.yml`, as they are already supported in DAG config.


pankajastro (approved):
Nice!

PR #457, opened 2025-07-04
Validate DAG's on Airflow 3

pankajastro:


.github/workflows/cicd.yaml
```diff
@@ -3,7 +3,7 @@ name: CI jobs
 on:
   # run on pushes to the default branch
   push:
-    branches: [main]
+    branches: [main, validate_operators]
```
```suggestion
    branches: [main]
```

.github/workflows/cicd.yaml
```diff
@@ -185,7 +185,7 @@ jobs:
           AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT: 90.0
           AIRFLOW_HOME: ${{ github.workspace }}
           CONFIG_ROOT_DIR: ${{ github.workspace }}/dags
-          PYTHONPATH: ${{ github.workspace }}:${{ github.workspace }}/examples:$PYTHONPATH
+          PYTHONPATH: ${{ github.workspace }}:${{ github.workspace }}/dev/dags:$PYTHONPATH
           AUTO_CONVERT_TO_AF3: true
```
tatiana: Should we remove this (or other references) now?

pankajastro: Some unit tests still reference import paths and argument names, such as airflow.operator.bash.BashOperator and schedule_interval. We need these since we run tests for Airflow 2. Similarly, for integration tests, we have configurations in dev/dags that are used for both Airflow 2 and Airflow 3, and they require conversion at runtime

pankajastro:
> Thank you very much for addressing the feedback and for your detailed comment, @pankajastro!
> 
> Do we have a ticket for removing the need to run `AUTO_CONVERT_TO_AF3` to run the dags in `dev/dags`?

Thanks! Created an issue: https://github.com/astronomer/dag-factory/issues/476 to remove `AUTO_CONVERT_TO_AF3`.  

>  I wonder if we could have version-agnostic DAGs in `dev/dags` and have the DAGs with specific AF2 & AF3 configurations in dedicated folders.

I think this change would make it more readable. Created issue: https://github.com/astronomer/dag-factory/issues/477 to improve it.

PR #444, opened 2025-06-27
Add support for custom type

pankajastro:
Add support for custom type. This implementation is an alternate solution for https://github.com/astronomer/dag-factory/pull/433

Example Config:
```
custom_example_dag:
  description: "this is an example dag"
  schedule_interval: "@daily"
  catchup: false
  render_template_as_native_obj: True
  tasks:
    task_1:
      operator: airflow.operators.bash_operator.BashOperator
      bash_command: "echo 1"
      executor_config:
        pod_override:
          __type__: kubernetes.client.models.V1Pod
          spec:
            __type__: kubernetes.client.models.V1PodSpec
            containers:
              __type__: builtins.list
              items:
                - __type__: kubernetes.client.models.V1Container
                  name: "base"
                  resources:
                    __type__: kubernetes.client.models.V1ResourceRequirements
                    limits:
                      cpu: "1"
                      memory: "1024Mi"
                    requests:
                      cpu: "0.5"
                      memory: "512Mi"
    task_2:
      operator: airflow.operators.bash_operator.BashOperator
      bash_command: "echo 2"
      dependencies: [task_1]
    task_3:
      operator: airflow.operators.bash_operator.BashOperator
      bash_command: "echo 2"
      dependencies: [task_1]
```
Screenshot 2025-06-28 at 1 32 04 AM

docs: https://astronomer.github.io/dag-factory/dev/features/custom_python_object/

closes: https://github.com/astronomer/dag-factory/issues/426
closes: https://github.com/astronomer/dag-factory/issues/376
closes: https://github.com/astronomer/dag-factory/issues/417
closes: https://github.com/astronomer/dag-factory/issues/104

PR #458, opened 2025-07-06
Refactor schedule to use the Python object

pankajastro:
This PR refactor the scheduling to take advantage of the custom Python object introduced in PR: https://github.com/astronomer/dag-factory/pull/444

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -811,7 +811,10 @@ class DagBuilder:
             if DagBuilder._is_asset(schedule):
                 dag_kwargs["schedule"] = DagBuilder._asset_schedule(schedule)
             else:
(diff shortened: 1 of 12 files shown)


tatiana (approved):
@pankajastro it makes sense to have a unified and consistent approach to typing in DAG Factory, thank you for proposing this.

I'd love users feedback on this interface, let's see if we get any during the next minor release - we can still review the strategy before the 1.0 release if we see the need for it.

pankajastro:
> @pankajastro it makes sense to have a unified and consistent approach to typing in DAG Factory, thank you for proposing this.
> 
> I'd love users feedback on this interface, let's see if we get any during the next minor release - we can still review the strategy before the 1.0 release if we see the need for it.


Sure, we can review and improve it based on user feedback. Thanks!

Issue #462, opened 2025-07-07
[Feature] additional DAG-level defaults beyond default_args


wearpants:
### Description

It'd be nice to be able to specify defaults for other DAG-level parameters in addition to `default_args` for tasks. I'm thinking things like `start_date`, `catchup`, `max_active_runs`, `params` etc..

### Use case/motivation

These parameters are often confusing for users and I like to apply a standard set of behaviors (for example, I always use `catchup=False`, etc.)

### Are you willing to submit a PR?

- [ ] Yes, I am willing to submit a PR!

gyli:
+1

PR #480, opened 2025-07-13
Support dag-level arguments in global defaults

gyli:
closes: https://github.com/astronomer/dag-factory/issues/462

### DIFF
--- a/tests/test_dagfactory.py
+++ b/tests/test_dagfactory.py
@@ -491,6 +491,49 @@ def test_build_dag_with_global_default():
     assert dags.get("example_dag").tasks[0].depends_on_past == True
 
 
+def test_build_dag_with_global_dag_level_defaults():
+    """Test that DAG-level defaults from global defaults.yml are applied to individual DAG configs"""
+    global_defaults = {
+        "default_args": {"owner": "global_owner"},
+        "schedule_interval": "0 0 * * *",
+        "catchup": False,
+        "tags": ["global_tag"]
+    }
+
+    config = {
+        "test_dag": {
+            "tasks": {
+                "task_1": {
+                    "operator": get_bash_operator_path(),
+                    "bash_command": "echo 1"
+                }
+            }
+        },
+        "test_dag_override": {
+            "schedule_interval": "0 1 * * *",
+            "tasks": {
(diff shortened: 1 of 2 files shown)


gyli:
Unit test fixed and doc added

tests/test_dagfactory.py
```diff
@@ -490,6 +490,48 @@ def test_build_dag_with_global_default():
     assert dags.get("example_dag").tasks[0].depends_on_past == True
 
 
+def test_build_dag_with_global_dag_level_defaults():
+    """Test that DAG-level defaults from global defaults.yml are applied to individual DAG configs"""
+    global_defaults = {
+        "default_args": {"owner": "global_owner", "start_date": "2020-01-01",},
+        "schedule": "0 1 * * *",
```
pankajastro:
```suggestion
        "schedule_interval": "0 1 * * *",
```

tests/test_dagfactory.py
```diff
@@ -490,6 +490,48 @@ def test_build_dag_with_global_default():
     assert dags.get("example_dag").tasks[0].depends_on_past == True
 
 
+def test_build_dag_with_global_dag_level_defaults():
+    """Test that DAG-level defaults from global defaults.yml are applied to individual DAG configs"""
+    global_defaults = {
+        "default_args": {"owner": "global_owner", "start_date": "2020-01-01",},
+        "schedule_interval": "0 1 * * *",
```
pankajastro:
```suggestion
        get_schedule_key(): "0 1 * * *",
```

tests/test_dagfactory.py
gyli: I've tried this, but it failed when testing with Airflow 3
```
dagfactory.exceptions.DagFactoryException: Failed to generate dag test_dag: 'schedule'
```

I believe it's because this test mock the output of `_global_default_args`, which contains `update_yaml_structure` updating config from Airflow 2 to 3. Hence it should be Airflow 3 keys used in this mock config.

pankajastro: We have added a few utilities like `get_schedule_key()`  to handle compatibility in the test.

PR #482, opened 2025-07-16
Add Zizmor analysis GitHub action

pankajkoti:
Adds an automated security analysis workflow that runs the Zizmor scanner on every push to main and on all pull requests.

related: https://github.com/astronomer/oss-integrations-private/issues/156

Reference: https://docs.zizmor.sh/

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,0 +1,24 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["**"]
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: Run zizmor 🌈
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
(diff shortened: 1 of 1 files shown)


pankajastro:
I'm just wondering if this job by default, uploads the scan artifact. I can see in the log `Uploading combined SARIF debug artifact`, but it is empty https://github.com/astronomer/dag-factory/actions/runs/16320884127/job/46098049719#step:5:3

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,24 @@
...
+    branches: ["**"]
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: Run zizmor 🌈
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
```
pankajastro, line 16: If we do not upload the scan result this permission may not require

pankajkoti: I believe it uploads the scan report and hence they appear here: https://github.com/astronomer/dag-factory/security/code-scanning?query=pr%3A482+is%3Aopen

PR #483, opened 2025-07-16
Restore the basic dag example

pankajastro:
Restore the basic example DAG

The below is incorrect since the file has moved a different location

> Create a file named example_dag_factory.yml in the $AIRFLOW_HOME/dags folder with the following content:
https://astronomer.github.io/dag-factory/latest/getting-started/quick-start-airflow-standalone/#step-5-define-a-dag-in-yaml

The current Python file does not look that good example for getting started
https://astronomer.github.io/dag-factory/latest/getting-started/quick-start-airflow-standalone/#step-6-generate-the-dag-from-yaml

reported: https://astronomer.slack.com/archives/C07CNMS6TKL/p1752627911018709


.github/workflows/cicd.yaml
```diff
@@ -3,7 +3,7 @@ name: CI jobs
 on:
   # run on pushes to the default branch
   push:
-    branches: [main]
+    branches: [main, restore_basic_dag]
```
```suggestion
    branches: [main]
```

.gitignore
```diff
@@ -125,7 +125,6 @@ logs/
 # Airflow
 examples/.airflowignore
 airflow.cfg
-dags
```
pankajkoti: is this removal intentional? probably this was here for symlinks? and need to be kept in gitignore?

pankajkoti (approved):
LGTM. Quick question in-line about the change in gitignore file

.gitignore
```diff
@@ -125,7 +125,6 @@ logs/
 # Airflow
 examples/.airflowignore
 airflow.cfg
```
pankajastro:
```suggestion
airflow.cfg
dags
```

PR #485, opened 2025-07-21
 Fix the Airflow version condition check to parse inlets/outlets syntax according to the dataset

pankajastro:
We should be able to parse inlets/outlets syntax according to the dataset for Airflow < 3.0.0, because in Airflow 3, the dataset has been replaced with an asset. This PR fixes the wrong condition check for Airflow version for parsing in inlets/outlets for datasets

This PR also add an example DAG for Asset.

Before this PR, DAG below will not work.

```yaml
default:
  default_args:
    start_date: 2025-01-01
  catchup: false

producer_dag:
  schedule: "@daily"
  tasks:
    produce_data:
      operator: "airflow.providers.standard.operators.python.PythonOperator"
      python_callable: utils.generate_data
      outlets:
        - __type__: airflow.sdk.Asset
          uri: "file:///$AIRFLOW_HONE/data.csv"
          name: "data_asset"

consumer_dag:
  schedule:
    - __type__: airflow.sdk.Asset
      uri: "file:///$AIRFLOW_HONE/data.csv"
      name: "data_asset"
  tasks:
    consume_data:
      operator: "airflow.providers.standard.operators.bash.BashOperator"
      bash_command: "echo 'Asset was updated, running DAG!'"
```


dagfactory/dagbuilder.py
```diff
@@ -334,6 +334,7 @@ def make_task(operator: str, task_params: Dict[str, Any]) -> BaseOperator:
 
         :returns: instance of operator object
         """
+        print("task_params: ", task_params)
```
pankajastro:
```suggestion
```

dagfactory/dagbuilder.py
```diff
@@ -955,7 +955,6 @@ def build(self) -> Dict[str, Union[str, DAG]]:
 
                 if task_conf.get("expand"):
                     task_conf = self.replace_expand_values(task_conf, tasks_dict)
```
pankajastro:
```suggestion
                if task_conf.get("expand"):
                    task_conf = self.replace_expand_values(task_conf, tasks_dict)

```

scripts/test/pre-install-airflow.sh
```diff
@@ -33,7 +33,7 @@ uv pip install pip --upgrade
 
 
 if [ "$AIRFLOW_VERSION" = "3.0" ]; then
-  uv pip install "apache-airflow>=3.0.2" --constraint /tmp/constraint.txt
+  uv pip install "apache-airflow==3.0.2" --constraint /tmp/constraint.txt
```
tatiana: Would it make sense to install "apache-airflow~=3.0.2"?

dev/dags/asset_triggered_dags.py
```diff
@@ -0,0 +1,17 @@
...
+
+DEFAULT_CONFIG_ROOT_DIR = "/usr/local/airflow/dags/"
+CONFIG_ROOT_DIR = Path(os.getenv("CONFIG_ROOT_DIR", DEFAULT_CONFIG_ROOT_DIR))
+
+config_file = str(CONFIG_ROOT_DIR / "asset_triggered_dags.yml")
+
+example_dag_factory = dagfactory.DagFactory(config_file)
+
+# Creating task dependencies
+example_dag_factory.clean_dags(globals())
```
pankajastro:
```suggestion
```

scripts/test/pre-install-airflow.sh
pankajastro:
```suggestion
  uv pip install "apache-airflow~=3.0.2" --constraint /tmp/constraint.txt
```

pankajastro: Fixed in 9a02238e3374fb6dc6702e84fa824657953aabb0

scripts/test/pre-install-airflow.sh
```diff
@@ -33,7 +33,7 @@ uv pip install pip --upgrade
 
 
 if [ "$AIRFLOW_VERSION" = "3.0" ]; then
-  uv pip install "apache-airflow>=3.0.2" --constraint /tmp/constraint.txt
+  uv pip install "apache-airflow~=3.0.2" --constraint /tmp/constraint.txt
```
pankajastro:
```suggestion
  uv pip install "apache-airflow>=3.0.2" --constraint /tmp/constraint.txt
```

pankajkoti (commented):
Sorry, I’m not sure what we’re trying to accomplish here. Could you link the relevant issue for context?

The PR description says, “We should parse inlets/outlets syntax as per the dataset for Airflow < 2.4,” which I find a bit confusing.

pankajastro:
> Sorry, I’m not sure what we’re trying to accomplish here. Could you link the relevant issue for context?
> 
> The PR description says, “We should parse inlets/outlets syntax as per the dataset for Airflow < 2.4,” which I find a bit confusing.

My bad. I have improved the PR description. Please let me know if there is any further confusion.

dev/dags/asset_triggered_dags.yml
```diff
@@ -0,0 +1,25 @@
+default:
```
pankajkoti, line 1: Were we able to test with Airflow standalone or Astro deployment that this dag works well since the CI is disabling to run this DAG?

pankajkoti (approved):
I am good on this one if we have tested the DAG being added in the PR (at least manually) with Airflow.

dev/dags/asset_triggered_dags.yml
pankajastro: Yes, worked fine with astro-cli.

in our repo, `make docker-run`

Issue #409, opened 2025-05-27
Make `apache-airflow-providers-cncf-kubernetes` an optional dependency


pankajastro:
Currently, `dag-factory` installs provider dependencies (e.g., apache-airflow-providers-cncf-kubernetes) even if the user is not utilising the related operator. This can lead to unnecessary bloat and conflicts in environments where those providers are not needed.

Make these provider dependencies optional, so they are only installed when explicitly required by the user.

Address: https://github.com/astronomer/dag-factory/pull/407#pullrequestreview-2869263750

PR #486, opened 2025-07-21
Make providers an optional dependency

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/409

- Guard the import for SqlSensor, HttpOperator, HttpSensor KPO and other Kubernetes import 
- If Package is not installed, log message about it
- Add comments that will help in future cleanup if we upgrade packages
- Introduce additional installation target, dag-factory[all, kubernetes, http, common-sql]

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -384,7 +384,6 @@ class DagBuilder:
                 )
             else:
                 response_check_name = task_params.pop("response_check_lambda", None)
-
                 task_params["response_check"]: Callable = utils.get_python_callable_lambda(response_check_name)
         return task_params
 
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import ast
 import inspect
+import logging
 import os
 import re
 import warnings
@@ -13,6 +14,7 @@ from functools import partial, reduce
 from typing import Any, Callable, Dict, List, Tuple, Union
 
 from airflow import configuration
+from packaging import version
 
 try:
     from airflow.sdk.bases.operator import BaseOperator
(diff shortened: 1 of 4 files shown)


dev/dags/example_http_operator_task.py
```diff
@@ -17,8 +26,10 @@
 CONFIG_ROOT_DIR = Path(os.getenv("CONFIG_ROOT_DIR", DEFAULT_CONFIG_ROOT_DIR))
 if HTTP_OPERATOR_AVAILABLE:
     config_file = str(CONFIG_ROOT_DIR / "example_http_operator_task.yml")
-else:
+elif SIMPLE_HTTP_OPERATOR_AVAILABLE:
     config_file = str(CONFIG_ROOT_DIR / "example_simple_http_operator_task.yml")
+else:
+    raise Exception("Package apache-airflow-providers-http is not installed.")
```
tatiana: Shall we use a more specific exception? It could be `DagFactoryException` as well
```suggestion
    raise ImportError("Package apache-airflow-providers-http is not installed.")
```

dev/dags/example_http_operator_task.py
```diff
@@ -5,8 +5,17 @@
...
-    HTTP_OPERATOR_AVAILABLE = False
+    try:
+        from airflow.providers.http.operators.http import SimpleHttpOperator  # noqa: F401
+
+        SIMPLE_HTTP_OPERATOR_AVAILABLE = True
+        HTTP_OPERATOR_AVAILABLE = False
+    except ImportError:
+        HTTP_OPERATOR_AVAILABLE = False
+        SIMPLE_HTTP_OPERATOR_AVAILABLE = False
+        raise Exception("Package apache-airflow-providers-http is not installed.")
```
tatiana: Related to https://github.com/astronomer/dag-factory/pull/486/files#r2228077442
```suggestion
        raise ImportError("Package apache-airflow-providers-http is not installed.")
```

dagfactory/dagbuilder.py
```diff
@@ -324,6 +351,45 @@ def _clean_kpo_task_params(task_params: dict) -> dict:
...
+                    or `response_check_lambda` parameter."
+                )
+            if task_params.get("response_check_file"):
+                task_params["response_check"]: Callable = utils.get_python_callable(
+                    task_params["response_check_name"], task_params["response_check_file"]
+                )
+                # remove dag-factory specific parameters
+                # Airflow 2.0 doesn't allow these to be passed to operator
+                del task_params["response_check_name"]
+                del task_params["response_check_file"]
```
tatiana: We may be able to simplify the code by doing something similar to this:

```suggestion
            # remove dag-factory specific parameters
            # Airflow 2.0 doesn't allow these to 
            response_check_name = task_params.pop("response_check_name", None)
            response_check_file = task_params.pop("response_check_file", None)
            if response_check_name
                task_params["response_check"]: Callable = utils.get_python_callable(
                    response_check_name, response_check_file
                )
```

This suggestion will not work out of the box because we're in the middle of an if-else statement, but the relevant part is to leverage `pop`.

tatiana (approved):
Looks great, @pankajastro , thank you very much!  It is great that we're making the K8s provider become an optional dependency. I left minor feedback inline.

Issue #468, opened 2025-07-09
Make tasks a list


pankajastro:
Tasks as a list seem relatively cleaner apprach

```
# file name: basic_example_dag.dag.yaml
default_args:
  owner: "custom_owner"
description: "this is an example dag"
schedule: "0 3 * * *"
tasks:
  - task_id: task_1
    operator: airflow.operators.bash_operator.BashOperator
    bash_command: "echo 1"
  - task_id: task_2
    operator: airflow.operators.bash_operator.BashOperator
    bash_command: "echo 2"
    dependencies:
	    - task_1
  - task_id: task_3
    operator: airflow.operators.bash_operator.BashOperator
    bash_command: "echo 2"
    dependencies:
	    - task_1
```

PR #487, opened 2025-07-22
Support tasks and task_groups as lists

pankajkoti:
### What?
This change adds first-class support for list-based syntax in DAG-Factory YAML:
1. Tasks as a list
2. Task Groups as a list

Until now, both sections had to be dictionaries keyed by _task_id_ / _group_id_. Authors can now opt for a cleaner, order-preserving list format while everything still works in dictionary form.

### Why?
The list style is more natural, readable, and aligns with common YAML patterns.

Example before ⬇️
```
tasks:
  task_1:
    bash_command: "echo 1"
```
Example after ⬇️
```
tasks:
  - task_id: task_1
    bash_command: "echo 1"
```

Same goes for `task_groups`:
```
task_groups:
  - group_name: extract
    tooltip: "extraction steps"
```

### Implementation details

`DagBuilder.build()` now:
Detects if `tasks` or `task_groups` is a list.
Converts each list to the existing dict structure (task_id / group_name key becomes dict key).
Performs validation:
1. Each entry must be a mapping with the required identifier field.
2. Duplicate identifiers raise DagFactoryConfigException.
No downstream code had to change – all later logic still sees a dictionary.

### Backwards compatibility
Existing dictionary syntax continues to work unchanged.
Tests covering legacy pattern remain green.

closes: #468

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -822,6 +822,85 @@ class DagBuilder:
                 else:
                     dag_kwargs["schedule"] = schedule
 
+    @staticmethod
+    def _normalise_tasks_config(tasks_cfg: Any) -> Dict[str, Dict[str, Any]]:
+        """Ensure tasks configuration is in the canonical dict form.
+
+        Dag authors may provide tasks either as a mapping of ``task_id`` -> config
+        or as a *list* of configs each containing a ``task_id`` key. This helper
+        converts the latter to the former so that the rest of the builder logic
+        can operate on a single, predictable structure.
+
+        :param tasks_cfg: the raw ``tasks`` value from the YAML / dict config
+        """
+        # Nothing provided – let the caller decide how to handle later.
+        if tasks_cfg is None:
+            return {}
+
+        # Already in the desired form
+        if isinstance(tasks_cfg, dict):
+            return tasks_cfg
+
+        if isinstance(tasks_cfg, list):
+            converted: Dict[str, Dict[str, Any]] = {}
+
+            for entry in tasks_cfg:
+                if not isinstance(entry, dict) or "task_id" not in entry:
+                    raise DagFactoryConfigException(
+                        "Each task definition in the list must be a mapping that contains a 'task_id' key"
+                    )
+
+                task_id = entry["task_id"]
+
+                if task_id in converted:
+                    raise DagFactoryConfigException(f"Duplicate task_id detected in tasks list: '{task_id}'")
+
+                # Exclude task_id from the configuration body – historically it
+                # is represented by the mapping key, not within the dict.
+                task_conf = {k: v for k, v in entry.items() if k != "task_id"}
+                converted[task_id] = task_conf
+
+            return converted
+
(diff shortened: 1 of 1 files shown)


dagfactory/dagbuilder.py
```diff
@@ -832,6 +832,70 @@ def build(self) -> Dict[str, Union[str, DAG]]:
...
+                task_conf = {k: v for k, v in task_entry.items() if k != "task_id"}
+                converted_tasks_cfg[task_id] = task_conf
+
+            # Replace the list with the converted dictionary so the rest of
+            # the build logic can stay exactly the same.
+            dag_params["tasks"] = converted_tasks_cfg
+
+        # ------------------------------------------------------------------
+        # Support configuration where ``task_groups`` can also be provided as
+        # a *list* of mappings – each with a mandatory ``group_id``.
```
pankajkoti:
```suggestion
        # a *list* of mappings – each with a mandatory ``group_name``.
```

tests/test_tasks_and_task_groups_list.py
```diff
@@ -0,0 +1,200 @@
...
+    dag_config["tasks"][0].pop("task_id")
+
+    td = DagBuilder("dag_missing_task_id", dag_config, DEFAULT_CONFIG_MINIMAL)
+    with pytest.raises(DagFactoryConfigException, match="task_id"):
+        td.build()
+
+
+def test_missing_task_group_name_raises_exception():
+    dag_config = _make_task_groups_list_config()
+    # Remove group_id from first task_group entry
```
pankajkoti:
```suggestion
    # Remove group_name from first task_group entry
```

pankajkoti:
I have logged a follow-up ticket: https://github.com/astronomer/dag-factory/issues/491 to update example DAGs and docs to reflect authoring tasks as list. Will work on a quick follow-up PR to do that once this change is avaialble and merged. Keeping the scope of this PR limited to ease reviews.

dagfactory/dagbuilder.py
```diff
@@ -832,6 +832,70 @@ def build(self) -> Dict[str, Union[str, DAG]]:
         """
         dag_params: Dict[str, Any] = self.get_dag_params()
 
+        # ------------------------------------------------------------------
```
tatiana: @pankajkoti previously, the `build` function was already quite big (~250 lines), would it be possible for us to break down the new logic in one or multiple functions, to help with the code maintainability?

Potentially, each block you commented in advance could become a function, and those comments may be docstring candidates.

tatiana (approved):
This is a very exciting feature, and it's great it is backwards compatible.
I left an implementation feedback inline, but happy with the feature!

dagfactory/dagbuilder.py
pankajkoti: Thanks @tatiana . Indeed an important suggestion. I have addressed it in https://github.com/astronomer/dag-factory/pull/487/commits/bdbf8df8f5c0e104e32727ae191094dcdfe95672

Issue #491, opened 2025-07-24
Update example DAGs to reflect tasks as list


pankajkoti:
Once PR #487 is merged, work on changing the existing example docs and references in docs to reflect authoring tasks as a list

PR #492, opened 2025-07-24
Replace the usages in example dags, tests and docs for tasks and taskgroups to be list

pankajkoti:
Following up on PR #487, this PR demonstrates and replaces the usages of tasks and task groups across all example dags, tests and static yamls in docs to be a list.

closes: #491

### DIFF
--- a/tests/test_dagfactory.py
+++ b/tests/test_dagfactory.py
@@ -754,3 +754,80 @@ def test_default_override_based_on_directory_tree(serialize_config_md_mock, tmp_
     assert default_config["a_param"] == "a"
     assert default_config["b_param"] == "b"
     assert default_config["c_param"] == "c"
+
+
+# Test to ensure backward compatibility for dictionary-style tasks and task_groups definitions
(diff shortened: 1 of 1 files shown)


tests/test_tasks_and_task_groups_list.py
```diff
@@ -1,200 +0,0 @@
-import datetime
```
tatiana: @pankajkoti, why did we remove this test file?

pankajkoti: since we modified out example DAGs and tests to have tasks as list, we have it covered testing tasks and taskgroup as list. But having a separate module to explicitly test that should not hurt, so I have re-added that module.

pankajkoti:
Can we please review this? I will work on a follow-up PR to demonstrate in docs with example dag that we support dictionary based approach too.

tatiana (approved):
Thanks a lot, @pankajkoti , happy to see the dictionary examples in a follow up PR!

PR #498, opened 2025-07-24
Remove `clean_dags` function

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/471

Automatic cleanup occurred when the DAG was refreshed. This happens at the frequency defined by TAIRFLOW__DAG_PROCESSOR__REFRESH_INTERVAL


tatiana (approved):
Thanks, @pankajastro !

Issue #469, opened 2025-07-09
Support for defaults.yaml files at the global and subdirectory level


pankajastro:
Introduce support for defaults.yaml files at the global and subdirectory level
Use predictable merge behaviour where the lowest level wins. Keep the default section in-DAG for fine-grained overrides and as a template pattern for scoped DAGs Use additive merge for tags and deep merge for default_args

- [x] default.yml currently only accepts [DAG default_args](https://github.com/apache/airflow/blob/f896807026c7f75f686309293e72a1140a7a9cf9/task-sdk/src/airflow/sdk/definitions/dag.py#L383).
- [x] Allow default.yml per folder and merge it with the nearest one will get more priority

tatiana:
The following PRs should cover this change:
* https://github.com/astronomer/dag-factory/pull/500
* https://github.com/astronomer/dag-factory/pull/504

Issue #472, opened 2025-07-09
Evaluate and Remove`DagFactory.generate_dag()`


pankajastro:

tatiana:
@pankajastro we may want to do #505 as part of this ticket

PR #500, opened 2025-07-25
Support overriding `defaults.yml` based on the directory hierarchy

tatiana:
Introduce support to multiple `defaults.yaml` files, allowing properties to be overridden following the proximity to the YAML DAG definition.

Use predictable merge behaviour where the level closest to the DAG YAML files wins. Keep the default section in-DAG for fine-grained overrides and as a template pattern for scoped DAGs.

- Added test that confirms `defaults.yml` not only accepts `default_args`, but also DAG-level arguments;
- Allow `defaults.yml` per folder, and merging it with the nearest one will get more priority;
- Merge tags from all `defaults.yml` configurations merged.

Closes: #469

The documentation was added in #504.

There are some open questions:
* Since `default_args_config_path` and `default_args_config_dict` support overriding not only `default_args`, but also any DAG-level args, we should rename them. I've logged a follow-up ticket: #505
* Currently `default_args_config_path` and `default_args_config_dict`  are exposed to users that leverage `DagFactory(...).generate_dags()`. If we intend to [deprecate this method](https://github.com/astronomer/dag-factory/issues/472), should we expose these fields via `load_yaml_dags`?


tests/fixtures_without_default/dag_factory.yml
```diff
@@ -0,0 +1,69 @@
+default:
```
pankajkoti: slightly confused with the placement of this file. The directory is `fixtures_without_default` and the file contains a default block.

tests/fixtures_without_default/dag_factory.yml
```diff
@@ -0,0 +1,69 @@
...
+  schedule_interval: 0 1 * * *
+example_dag:
+  default_args:
+    owner: custom_owner
+    start_date: 2 days
+  description: this is an example dag
+  doc_md: '##here is a doc md string'
+  schedule_interval: 0 3 * * *
+  tasks:
+    task_1:
```
pankajkoti: Could we please represent the tasks as a list here since we now have PR #487 merged? There is a PR #492 attempting to convert all those tasks to a list in our example DAGs. So it would be nice if any new examples we add have them as lists.

pankajkoti (approved):
LGTM. Happy with the implementation. Coupld of in-line comments (a quick question and a suggestion).

tests/fixtures_without_default/dag_factory.yml
tatiana: It's a fixtures directory without a `defaults.yaml` file 🙈 
Would it be clearer if I renamed to `fixtures_without_default_yaml`?

tests/fixtures_without_default/dag_factory.yml
tatiana: Sure, this was copied from a previous one in the other directory, I'll copy the latest merged one!

tatiana: Addressed in 5d01272a13cc636ab4682d2ebc958013b104d7af

tests/fixtures_without_default/dag_factory.yml
pankajkoti: yes, good idea. probably that could help

pankajastro (approved):

tests/fixtures_without_default/dag_factory.yml
tatiana: Changed in 2fd9c757d564eed1686f24edf353d166db518d64

PR #504, opened 2025-07-25
Update default documentation based on #500

tatiana:
The PR #500 adds support for overriding `defaults.yml` based on the directory hierarchy.

This PR aims to review and document all options for defining default arguments in DAG Factory following this change.

### DIFF
--- a/docs/configuration/defaults.md
+++ b/docs/configuration/defaults.md
@@ -1,80 +1,105 @@
 # Defaults
 
-DAG Factory allows you to define Airflow
-[default_args](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#default-arguments) and
-additional DAG-level arguments in a `default` block. This block enables you to share common settings and configurations
-across all DAGs in your YAML configuration, with the arguments automatically applied to each DAG defined in the file.
-This is one of DAG Factory's most powerful features; using defaults allows for the dynamic generation of more than a
-single DAG.
+DAG Factory allows you to define default values for DAG-level arguments and Airflow
+[default_args](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#default-arguments). There are several ways to accomplish this:
 
-## Benefits of using the default block
+- Define the `default_args` within each DAG definition in the DAGs YAML file;
+- Declare a `default` block within the toplevel of the DAGs YAML file;
+- Define the `default_args_config_dict` argument when instantiating the `DAGFactory` class;
+- Create one or multiple `defaults.yml` and declare the `default_args_config_path` argument in the `DAGFactory` class. This approach includes support for combining multiple `defaults.yml` files.
+
+Although you cannot use the last two configurations together, you can use a combination of the first two configurations with either the third or the last.
+
+Below, we detail how to use each of these approaches and also how to combine them.
+
+## Specifying `default_args` directly in the DAG YAML specification
+
+This configuration affects only the DAG where the `default_args` are defined.
+
+You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but also adding new arguments new arguments.
+
+```yaml
+etl:
+   default_args:
+      start_date: '2024-12-31'
+      retries: 1  # A new default_arg was added
+...
+```
+
+## YAML top-level `default`
+
+This configuration affects all the DAGs defined in the YAML file where the `default` block is declared.
+
+The `default` top-level block enables you to share standard settings and configurations across all DAGs in your YAML configuration, with the arguments automatically applied to each DAG defined in the file.
+It is one of DAG Factory's most powerful features; using defaults allows for the dynamic generation of multiple DAGs.
+
+### Benefits of using the `default` block
 
 - Consistency: Ensures uniform configurations across all tasks and DAGs.
 - Maintainability: Reduces duplication by centralizing common properties.
 - Simplicity: Makes configurations easier to read and manage.
-- Dynamic Generation: Use a single default block to easily generate more than a single DAG.
+- Dynamic Generation: Use a single default block to generate more than a single DAG easily.
 
-### Example usage of a default block for `default_args`
+### Example usage of the `default` block
 
 #### Specifying `default_args` in the `default` block
 
-   Using a `default` block in a YAML file allows for those key-value pairs to be applied to each DAG that is defined in
-   that same file. One of the most common examples is using a `default` block to specify `default_args` for each DAG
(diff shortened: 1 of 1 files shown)


docs/configuration/defaults.md
```diff
@@ -1,80 +1,105 @@
...
+
+Although you cannot use the last two configurations together, you can use a combination of the first two configurations with either the third or the last.
+
+Below, we detail how to use each of these approaches and also how to combine them.
+
+## Specifying `default_args` directly in the DAG YAML specification
+
+This configuration affects only the DAG where the `default_args` are defined.
+
+You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but also adding new arguments new arguments.
```
pankajkoti:
```suggestion
You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but also adding new arguments.
```
Can we please re-check the phrasing. I did not fully understand what we meant on the first pass of read. The part I did not understand is where do we mean existing `default_args` would come from and also what do we mean about the new arguments and where would they come from.

docs/configuration/defaults.md
```diff
@@ -1,80 +1,105 @@
...
 
-   Starting DAG Factory 0.22.0, you can also keep the `default_args` in the `defaults.yml` file. The configuration
-   from `defaults.yml` will be applied to all DAG Factory generated DAGs. **Be careful, these will be applied to all
-   generated DAGs.**
+This configuration affects DAGs created using the `DagFactory` class with the `default_args_config_dict` argument.
 
-   ```yaml title="defaults.yml"
-   --8<-- "dev/dags/defaults.yml"
-   ```
+It allows you to define DAG-level arguments, included the `default_args`, using Python dictionaries.
```
pankajkoti:
```suggestion
It allows you to define DAG-level arguments, including the `default_args`, using Python dictionaries.
```

docs/configuration/defaults.md
```diff
@@ -85,16 +110,47 @@ single DAG.
...
+
+Assuming you instantiate the DAG by using:
+
+```python
+DagFactory(
+   "a/b/c/some_dags.yml",
+   default_args_config_path="a/b/c"
+)`
+
+The DAG will be using the default configuration defined in all the following files:
```
pankajkoti, line 140: I had a slightly different thought process here. I was thinking that we should not traverse up the path beyond what has been specified in `default_args_config_path`, but instead traverse down from this path until the DAG YAML directory (given the path is a parent directory to the DAG YAML and if not, just use that single path default without traversing). This way, we would also allow users to avoid using parent defaults by just overriding a defaults.yml closest to the DAG YAML. WDYT?

docs/configuration/defaults.md
```diff
@@ -85,16 +110,47 @@ single DAG.
...
+    │   └── defaults.yml
+    └── defaults.yml
 ```
+
+Assuming you instantiate the DAG by using:
+
+```python
+DagFactory(
+   "a/b/c/some_dags.yml",
+   default_args_config_path="a/b/c"
```
tatiana:
```suggestion
   default_args_config_path="a"
```

docs/configuration/defaults.md
tatiana: @pankajkoti your understanding is correct, and it was a documentation mistake, which I fixed in 616497acb137f7630a768ae7f68e04ae6821048e

docs/configuration/defaults.md
```diff
@@ -1,80 +1,105 @@
...
+
+Although you cannot use the last two configurations together, you can use a combination of the first two configurations with either the third or the last.
+
+Below, we detail how to use each of these approaches and also how to combine them.
+
+## Specifying `default_args` directly in the DAG YAML specification
+
+This configuration affects only the DAG where the `default_args` are defined.
+
+You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but also adding new arguments.
```
tatiana:
```suggestion
You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but also adding other DAG-level arguments.
```

docs/configuration/defaults.md
tatiana: @pankajkoti this paragraph was already existed and I just moved it around.

I tried to improve the description on 2db438d800928437add8a429ee38b6c4bfe95eea, please, let me know if you think it should be changed

docs/configuration/defaults.md
```diff
@@ -1,80 +1,105 @@
...
+
+Although you cannot use the last two configurations together, you can use a combination of the first two configurations with either the third or the last.
+
+Below, we detail how to use each of these approaches and also how to combine them.
+
+## Specifying `default_args` directly in the DAG YAML specification
+
+This configuration affects only the DAG where the `default_args` are defined.
+
+You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but also adding other DAG-level arguments.
```
pankajkoti:
```suggestion
You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but other DAG-level arguments can be added.
```

pankajkoti (approved):
LGTM. Thanks for the quick clarification!

pankajastro (approved):

docs/configuration/defaults.md
```diff
@@ -1,80 +1,105 @@
...
+
+Although you cannot use the last two configurations together, you can use a combination of the first two configurations with either the third or the last.
+
+Below, we detail how to use each of these approaches and also how to combine them.
+
+## Specifying `default_args` directly in the DAG YAML specification
+
+This configuration affects only the DAG where the `default_args` are defined.
+
+You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but other DAG-level arguments can be added.
```
pankajkoti:
```suggestion
You can override or define specific `default_args` at the individual DAG level. This strategy allows you to customize arguments for each DAG without affecting others. Not only can existing `default_args` be overridden directly in a DAG configuration, but also other DAG-level arguments can be added.
```

Sorry for another quick suggestion, we had the "also" there before, my previous suggestion removed that.

PR #506, opened 2025-07-25
Add more examples for the Custom Python object

pankajastro:
Add more examples for the Custom Python object


tatiana (approved):
Looks great, @pankajastro - it will help in the transition form `_sec` fields.

Issue #476, opened 2025-07-11
Remove `AUTO_CONVERT_TO_AF3` from CI


pankajastro:
TODO: @pankajastro add details

PR #509, opened 2025-07-28
Change  `DagFactory` class access to private 

### LINKED ISSUE #472: Evaluate and Remove`DagFactory.generate_dag()`

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/472

This PR disable the below interface to create a DAG with DAG-Factory
```python3
import dagfactory

example_dag_factory = dagfactory.DagFactory("path/to/config.yml")

# Creating task dependencies
example_dag_factory.generate_dags(globals())
```
Additionally, it unifies the interface into a single method for DAG creation with DAG-Factory

```python3
from dagfactory import load_yaml_dags

load_yaml_dags(
    globals_dict=globals(),
   ...
)
```

Previously, `load_yaml_dags(...)` only accepted a folder as the configuration input. This PR introduces the following parameters to support handling a single configuration file and dict input:

- config_filepath
- config_dict
- default_args_config_dict

The order in which these parameters will apply is as follows:

- If config_filepath is provided, config_filepath and default_args_config_filepath will be used.
- If config_dict is provided, config_dict and default_args_config_dict will be used.
- If neither of the above is provided, the function will iterate over dags_folder and render all DAGs (The current main behaviour if `load_dags(...)`)

Example for passing single yaml file
```python3
import os
from pathlib import Path

# The following import is here so Airflow parses this file
# from airflow import DAG
from dagfactory import load_yaml_dags

DEFAULT_CONFIG_ROOT_DIR = "/usr/local/airflow/dags/"

config_file = str(CONFIG_ROOT_DIR / "example_callbacks.yml")


load_yaml_dags(
    globals_dict=globals(),
    config_filepath=config_file,
)
```

Removed the import path

```python3
from dagfactory import DagFactory
```

### DIFF
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -23,7 +23,7 @@ from dagfactory.exceptions import DagFactoryConfigException, DagFactoryException
 SYSTEM_PARAMS: List[str] = ["default", "task_groups"]
 
 
-class DagFactory:
+class _DagFactory:
     """
     Takes a YAML config or a python dictionary and generates DAGs.
 
@@ -49,7 +49,7 @@ class DagFactory:
         assert bool(config_filepath) ^ bool(config), "Either `config_filepath` or `config` should be provided"
 
         if config_filepath:
-            DagFactory._validate_config_filepath(config_filepath=config_filepath)
+            _DagFactory._validate_config_filepath(config_filepath=config_filepath)
             self.config: Dict[str, Any] = self._load_dag_config(config_filepath=config_filepath)
         if config:
             self.config = config
@@ -304,10 +304,10 @@ def load_yaml_dags(
     candidate_dag_files = []
 
     if config_filepath:
-        factory = DagFactory(config_filepath=config_filepath, default_args_config_path=default_args_config_path)
+        factory = _DagFactory(config_filepath=config_filepath, default_args_config_path=default_args_config_path)
         factory._generate_dags(globals_dict)
     elif config_dict:
-        factory = DagFactory(config=config_dict, default_args_config_dict=default_args_config_dict)
+        factory = _DagFactory(config=config_dict, default_args_config_dict=default_args_config_dict)
         factory._generate_dags(globals_dict)
     else:
         for suf in suffix:
@@ -316,7 +316,7 @@ def load_yaml_dags(
             config_file_abs_path = str(config_file_path.absolute())
             logging.info("Loading %s", config_file_abs_path)
             try:
-                factory = DagFactory(config_file_abs_path, default_args_config_path=default_args_config_path)
+                factory = _DagFactory(config_file_abs_path, default_args_config_path=default_args_config_path)
                 factory._generate_dags(globals_dict)
             except Exception:  # pylint: disable=broad-except
                 logging.exception("Failed to load dag from %s", config_file_path)
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -195,9 +195,7 @@ class DagFactory:
 
         :returns: dict from YAML config file
         """
-        # pylint: disable=consider-using-with
-        config = load_yaml_file(config_filepath)
-        return config
+        return load_yaml_file(config_filepath)
 
     def get_dag_configs(self) -> Dict[str, Dict[str, Any]]:
         """
@@ -262,7 +260,7 @@ class DagFactory:
         for dag_id, dag in dags.items():
             globals[dag_id]: DAG = dag
 
-    def generate_dags(self, globals: Dict[str, Any]) -> None:
+    def _generate_dags(self, globals: Dict[str, Any]) -> None:
         """
         Generates DAGs from YAML config
 
@@ -277,10 +275,13 @@ def load_yaml_dags(
     globals_dict: Dict[str, Any],
     dags_folder: str = airflow_conf.get("core", "dags_folder"),
     default_args_config_path: str = airflow_conf.get("core", "dags_folder"),
+    config_filepath: Optional[str] = None,
+    config_dict: Optional[dict] = None,
+    default_args_config_dict: Optional[dict] = None,
     suffix=None,
 ):
     """
-    Loads all the yaml/yml files in the dags folder
+    Loads YAML or YML files in a specified folder (or from a specific YAML file or dictionary)
 
     The dags folder is defaulted to the airflow dags folder if unspecified.
     And the prefix is set to yaml/yml by default. However, it can be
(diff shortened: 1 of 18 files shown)


pankajkoti (commented):
The approach looks good to me. I guess there's no direct way in stopping users to still use them, but it would be nice if we could remove showcasing it in our example DAGs and documents to reduce visibility of that

Will take a re-review once we conclude on the approach and tests pass in the CI

dagfactory/dagfactory.py
```diff
@@ -74,6 +74,13 @@ def __init__(
         self.default_args_config_path: str = default_args_config_path
         self.default_args_config_dict: Optional[dict] = default_args_config_dict
 
+        # This will only invoke in the CI
```
pankajkoti: Why will this be only invoked in the CI? If someone has set the environment variable, it will also apply to them, right? Can we remove this comment line if it's here by mistake?

dev/dags/example_dag_factory_default_config_dict.yml
```diff
@@ -1,16 +0,0 @@
-daily_etl:
```
pankajkoti: Why are we dropping this file?

dev/dags/example_dag_factory_default_config_dict.py
```diff
@@ -1,18 +1,29 @@
...
+            "transform": {
+                "operator": "airflow.operators.bash.BashOperator",
+                "bash_command": "echo transform",
+                "dependencies": ["extract"],
+            },
+            "load": {
+                "operator": "airflow.operators.bash.BashOperator",
+                "bash_command": "echo load",
+                "dependencies": ["transform"],
+            },
```
pankajkoti: Would it be possible to modify these to have list of tasks notation?

dagfactory/dagfactory.py
```diff
@@ -304,22 +312,33 @@ def load_yaml_dags(
     :param globals_dict: The globals() from the file used to generate DAGs
     :param dags_folder: Path to the folder you want to get recursively scanned
     :param default_args_config_path: The Folder path where defaults.yml exist.
+    :param config_filepath: A YAML path for DAG config.
+    :param config_dict: The DAG dictionary.
+    :param default_args_config_dict: The dictionary that hold default value.
```
pankajkoti, line 295: Can we add a TODO comment here and link the ticket https://github.com/astronomer/dag-factory/issues/527 to support ignoring YAMLs files optionally if someone needs to.

pankajkoti (approved):
LGTM. Some questions and suggestions in-line.

tatiana (commented):
This is an interesting change. It seems the right direction to have a single entry-point to convert YAMLs into DAGs in DagFactory.

It's a shame we don't know how often people have been using the `dagfactory.DagFactory` and `load_yaml_dags`  interfaces, to make a more informed decision on which of them to chose. The one advantage I can see of `dagfactory.DagFactory` in comparison to `load_yaml_dags` is that people can subclass, override methods and customise specific behaviours of DagFactory. By only exposing `load_yaml_dags` they could potentially still override, but would have to monkey-patch.

Hopefully, if this is a common pattern or if there are any other reasons why people would use `dagfactory.DagFactory` instead of `load_yaml` they will share feedback - and we can review this decision.

It seems we missed to update a comment and two example DAGs to not use the generate_dags method:
```
./tests/test_dagbuilder_httpoperator.py:    # Call generate_dags to build all DAGs from the YAML file
./scripts/airflow3/dags/example_dag_factory.py:example_dag_factory.generate_dags(globals())
./dev/dags/asset_triggered_dags.py:example_dag_factory.generate_dags(globals())
```

dagfactory/dagfactory.py
```diff
@@ -304,22 +312,35 @@ def load_yaml_dags(
...
+        for config_file_path in candidate_dag_files:
+            config_file_abs_path = str(config_file_path.absolute())
+            logging.info("Loading %s", config_file_abs_path)
+            try:
+                factory = DagFactory(config_file_abs_path, default_args_config_path=default_args_config_path)
+                factory._generate_dags(globals_dict)
+            except Exception:  # pylint: disable=broad-except
+                logging.exception("Failed to load dag from %s", config_file_path)
+            else:
+                logging.info("DAG loaded: %s", config_file_path)
```
tatiana: @pankajastro I was a bit confused about this part, it feels a very similar logic to what iwas mplemented in #500 , in particular:
https://github.com/astronomer/dag-factory/blob/2fd9c757d564eed1686f24edf353d166db518d64/dagfactory/dagfactory.py#L43-L74

Why not just reuse what was available in the now internal `DagFactory` class, instantiating it with the necessary arguments? We shouldn't be duplicating this type of logic in multiple places.

tatiana (commented):
I'm very supportive of us having a single entry point to load DAGs in DagFactory.
It feels redundant to have both:
- `load_yaml_dags`
- `DagFactory.generate_dags`

The one advantage I can see with sticking to `DagFactory.generate_dags` is the fact that people could easily subclass and override behaviours, if needed. By only exposing `load_yaml_dags`, people would need to monkey-patch to override behaviours, which is not as nice.

At the same time, I can understand a method is a simpler entry point, simpler for - likely -  most users.

it is a shame we don't have any metrics to help us deciding on which interface to keep. I'm hopeful end-users will give us feedback and we can revisit / make this method public again if there is demand.

Would it make sense to make the `DagFactory` class private instead of just the method?

dagfactory/dagfactory.py
pankajastro: Added the GH issue here as comment

dagfactory/dagfactory.py
pankajastro: We have an issue to remove this env: https://github.com/astronomer/dag-factory/issues/476
extended comment to include this ticket

dev/dags/example_dag_factory_default_config_dict.yml
pankajastro: From the file name and how we are using it in the Python file, it feels that we are trying to use it as a dict and not yml config file. I have added this as a dict and removed the YAML file. See file: dev/dags/example_dag_factory_default_config_dict.py. Please let me know something looks odd

pankajastro:
> It seems we missed to update a comment and two example DAGs to not use the generate_dags method:
>
> ./tests/test_dagbuilder_httpoperator.py:    # Call generate_dags to build all DAGs from the YAML file
> ./scripts/airflow3/dags/example_dag_factory.py:example_dag_factory.generate_dags(globals())
> ./dev/dags/asset_triggered_dags.py:example_dag_factory.generate_dags(globals())

Yes, these DAG were recently added. Fixed them!

dev/dags/example_dag_factory_default_config_dict.py
pankajastro: yes, changed to list

dagfactory/dagfactory.py
pankajastro: Hi @tatiana, #500 merges the defaults.yml. Could you please point me to the function/method you are talking about?

tatiana: Ah, I see, `config_filepath` means YAML DAG filepath instead of "configuration" (that contains defaults) - thanks, @pankajastro for the clarification!

Issue #284, opened 2024-11-15
Improve DAG Factory nomenclature consistencies


tatiana:
This feedback was [reported](https://docs.google.com/document/d/1u_9Xn0U6y4hEz3GWmPA4D7c73U1yN_f89ACXHAxrq7w/edit?tab=t.0) by Karen Braganza:

> Some parameter names do not match between Airflow and DAG Factory (example: `schedule` vs `schedule_interval`  or `execution_timeout` vs `execution_timeout_secs`). This can be confusing without comprehensive documentation. I often had to read through the code to determine correct parameter names, which defeats the purpose of DAG factory for people unfamiliar with Python.
>
> This confusion is exacerbated by inconsistencies in nomenclature within DAG factory itself. Example: 
> * `retry_delay_sec` and `dagrun_timeout_sec` end with sec
> * On the other hand, `execution_timeout_secs` and `sla_secs` end with secs.
> * There is an extra ‘s’ at the end of `execution_timeout_secs` and `sla_secs`. These parameter names should be consistent.

I believe some of the inconsistencies may have been introduced for backwards compatibility since DAG Factory has historically supported many versions of Airflow, including 1.x. That said, I agree with Karen we should review and make these more consistent.

PR #512, opened 2025-07-28
Breaking Changes: Remove Inconsistent Parameters for Airflow Consistent

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/284

This PR removes the following parameters to make the DAG factory parameters consistent with Airflow parameters (Breaking changes)

- Removed `dagrun_timeout_sec` from dag param.
- Removed `retry_delay_sec`, `sla_secs` and `execution_timeout` from `default_args`.
- Removed `execution_timeout_secs` `sla_secs` and `execution_delta_secs` from task param

I have added more documentation in PR: https://github.com/astronomer/dag-factory/pull/506 for `__type__` uses.

### DIFF
--- a/tests/fixtures/dag_factory_kubernetes_pod_operator.yml
+++ b/tests/fixtures/dag_factory_kubernetes_pod_operator.yml
@@ -4,7 +4,7 @@ default:
     start_date: 2018-03-01
     end_date: 2018-03-05
     retries: 1
-    retry_delayc:
+    retry_delay:
       __type__: datetime.timedelta
       seconds: 300
   concurrency: 1


tatiana (commented):
@pankajastro Looks great, we can handle similar use-cases leveraging the recently introduced type casting #444.

It seems there are still some YAML DAGs that have attributes with the suffix `_sec`:
```
tests/fixtures/dag_factory_task_group.yml:    retry_delay_sec: 300
tests/fixtures/dag_factory.yml:  dagrun_timeout_sec: 600
tests/fixtures/dag_factory.yml:    retry_delay_sec: 300
```

And some pending tests as well:
```
tests/test_dagfactory.py:                "retry_delay_sec": 300,
tests/test_dagfactory.py:            "dagrun_timeout_sec": 600,
tests/test_dagfactory.py:            "retry_delay_sec": 300,
tests/test_dagfactory.py:        "dagrun_timeout_sec": 600,
tests/test_dagfactory.py:  dagrun_timeout_sec: 600
tests/test_dagfactory.py:    retry_delay_sec: 300
```

Do we want to update those as well?

tests/fixtures_without_default_yaml/dag_factory.yml
```diff
@@ -1,11 +1,9 @@
 default:
   concurrency: 1
-  dagrun_timeout_sec: 600
```
pankajastro, line 3: Removed this parameter because we convert the dictionary to a Python object internally and render it as such. Handling this for the md_docs test became a bit non-idiomatic. We have other tests in place to check these conversions.

```
dagrun_timeout: !!python/object/apply:datetime.timedelta
    - 0
    - 600
    - 0
```

tests/fixtures/dag_factory_kubernetes_pod_operator.yml
```diff
@@ -4,10 +4,14 @@ default:
     start_date: 2018-03-01
     end_date: 2018-03-05
     retries: 1
-    retry_delay_sec: 300
+    retry_delayc:
```
pankajkoti:
```suggestion
    retry_delay:
```

tests/fixtures/dag_factory.yml
```diff
@@ -1,11 +1,15 @@
 default:
   concurrency: 1
-  dagrun_timeout_sec: 600
+  dagrun_timeout:
+    __type__: datetime.timedelta
```
pankajkoti, line 4: Maybe we could make dag authoring simpler by accepting int values as the default getting converted to datetime.timedelta seconds, so that users won't have to specify a __type__ construct unless they want something different than the default. WDYT?

tests/fixtures/dag_factory.yml
```diff
@@ -1,11 +1,15 @@
...
+    seconds: 600
   default_args:
     end_date: 2018-03-05
     owner: default_owner
     retries: 1
-    retry_delay_sec: 300
+    retry_delay:
+      __type__: datetime.timedelta
+      seconds: 300
```
pankajkoti, line 12: would it be possible to assert in tests that dags generated from yaml essentially contain the desired type and values for these renamed keys?

pankajkoti (approved):
I am good on this, have a question inline about having default values for the keys without the need for specifying types. 

Important one, before merging, can we add assertion tests that generated dags from YAMLs contain the `dagrun_timeout` and `retry_delay` with desired type and values?

tests/fixtures/dag_factory.yml
pankajastro: Yes, good point. Added test: d1ca1c6a65d8a513731aeb7f1be409a5917cab49

tests/fixtures/dag_factory.yml
pankajastro: Maybe or maybe not. Created an issue: https://github.com/astronomer/dag-factory/issues/528 to discuss this

Issue #445, opened 2025-06-27
[Feature] DAG-factory CLI


gyli:
### Description

A CLI tool for DAG-factory can be created as the unified entry point for various operations, such as migrating config for Airflow 2 to 3, config linter, etc.

### Related issues

Airflow 2 to 3 config migration: https://github.com/astronomer/dag-factory/issues/439
Config linting: https://github.com/astronomer/dag-factory/issues/87

### Are you willing to submit a PR?

- [x] Yes, I am willing to submit a PR!

pankajastro:
I think this would make an excellent addition.

tatiana:
This work was stared in the following PRs:
1. #510
2. #511
3. #513

tatiana:
I'll soon add the `convert` command

Issue #87, opened 2021-07-02
Add json schema validation for yaml files


dinigo:
Hi there 👋 

When I write code I always use a linter, my IDE tells me when me when I'm about to make a mess by not importing certain library or not using the right type.

This does not apply to config files, They're super nice because config "always compiles" 😉 . But the code running those config files not always like it. So to have a little bit more control I try to implement "schemas" on those config files. They're looser than "clases", also language independent, easier to maintain and to run.

It would be nice to have something like

```python
import yaml
from jsonschema import validate

if __name__ == '__main__':
    schema = """
        type: object
        properties:
          default_args:
            type: object
            properties:
              start_date:
                type: string
          schedule_interval:
            type: string
          catchup:
            type: boolean
          tasks:
            type: array
            items:
              type: object
              properties:
                operator:
                  type: string
                dependencies:
                  type: array
                  items:
                    type: string
    """
    validate(
        instance=yaml.safe_load(open('path/to/my_dag.yml')),
        schema=yaml.safe_load(schema)
    )

But this `schema` is the one I generated, it doesn't distinguish between airflow versions, it's not up to date with the library. It would be nice to have a check like this. Pycharm picks up those schemas to validate things like  `.gitlab-ci.yaml`. So we could as well be writing DAGs the "safest" way with free linting.

Also a though. But this way the codebase could ease up a little bit on the yaml validation. Reducing the `core` codebase makes it easier to read and maintain.

Regardless of what you think of this porposal, thanks for your work, and for open sourcing this library. It rocks 🚀 !

ajbosco:
This looks like a great idea! I'm not sure I'll get to it for a bit though, but feel free to submit a PR!

gyli:
+1 on this one. Other than in-IDE checking, YAML validation in CI can be performed if there's such jsonschema, which could be very helpful to the YAML DAG development.

PR #513, opened 2025-07-28
Add CLI `lint` command

tatiana:
Introduce a lint command in the DAG Factory CLI.

There are two types of lint currently checked:
* Check if this is a valid YAML.
* Check if this is a DAG Factory compatible YAML (which has boolean operations and typing)

In future, this feature can be further improved by utilising a JSON Schema, as proposed in #87; however, it serves as a good starting point. The command can be used with or without the flag `--verbose`.

Related ticket: #445

**How this was tested**

Running

```bash
PYTHONPATH=`pwd` python -m dagfactory lint <path to directory or file>
```
Examples of invalid files:

```
# invalid-files/a.yml 
fruits:
  - [orange, mango
```

```
# invalid-files/b.yaml 
database
  host: localhost
```

**Screenshots**

Example running the lint command with success:
Screenshot 2025-07-28 at 19 48 44

Example running the lint command with invalid files:
Screenshot 2025-07-28 at 19 48 04

Example running the lint command using the `--verbose` flag:
Screenshot 2025-07-28 at 19 47 54

### DIFF
--- a/dagfactory/_yaml.py
+++ b/dagfactory/_yaml.py
@@ -1,0 +1,36 @@
+from __future__ import annotations
+
+import os
+
+import yaml
+
+from .utils import cast_with_type
+
+
+def load_yaml_file(file_path: str) -> dict[str, any]:
+    """
+    Load a YAML file into a dictionary.
+    """
+
+    def __join(loader: yaml.FullLoader, node: yaml.Node) -> str:
+        seq = loader.construct_sequence(node)
+        return "".join([str(i) for i in seq])
+
+    def __or(loader: yaml.FullLoader, node: yaml.Node) -> str:
+        seq = loader.construct_sequence(node)
+        return " | ".join([f"({str(i)})" for i in seq])
+
+    def __and(loader: yaml.FullLoader, node: yaml.Node) -> str:
+        seq = loader.construct_sequence(node)
+        return " & ".join([f"({str(i)})" for i in seq])
+
+    yaml.add_constructor("!join", __join, yaml.FullLoader)
+    yaml.add_constructor("!or", __or, yaml.FullLoader)
+    yaml.add_constructor("!and", __and, yaml.FullLoader)
+
+    with open(file_path, "r", encoding="utf-8") as fp:
+        config_with_env = os.path.expandvars(fp.read())
+        config: dict[str, any] = yaml.load(stream=config_with_env, Loader=yaml.FullLoader)
+        config = cast_with_type(config)
+
+    return config
--- a/docs/features/cli.md
+++ b/docs/features/cli.md
@@ -2,18 +2,62 @@
 
 After installing DAG Factory, the CLI can be invoked using the `dagfactory` command.
 
-## Usage
+## Commands summary
+
+| Command | Args   | Flags       | Description                                          |
+| ------- | ------ | ----------- | ---------------------------------------------------- |
+| `lint`  | `path` | `--verbose` | Check if the given directory or file is a valid YAML |
+
+For more details about the available commands, run `dagfactory --help`.
+
+### Example
+
+```bash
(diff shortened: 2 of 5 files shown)


docs/features/cli.md
```diff
@@ -2,18 +2,62 @@
...
 dagfactory [OPTIONS]
 ```
 
-| Option      | Alias | Description                                        |
+### Flags
+
+| Flag        | Alias | Description                                        |
 | ----------- | ----- | -------------------------------------------------- |
-| `--version` | `-v`  | Show the installed version of DAG Factory and exit |
+| `--version` |       | Show the installed version of DAG Factory and exit |
```
pankajastro, line 57: Why are you removing the -v option? Don’t we have an alias for --version?

dagfactory/_yaml.py
```diff
@@ -0,0 +1,36 @@
...
+        seq = loader.construct_sequence(node)
+        return " | ".join([f"({str(i)})" for i in seq])
+
+    def __and(loader: yaml.FullLoader, node: yaml.Node) -> str:
+        seq = loader.construct_sequence(node)
+        return " & ".join([f"({str(i)})" for i in seq])
+
+    yaml.add_constructor("!join", __join, yaml.FullLoader)
+    yaml.add_constructor("!or", __or, yaml.FullLoader)
+    yaml.add_constructor("!and", __and, yaml.FullLoader)
```
pankajastro, line 29: We may need to reconsider this because this custom tag may not work with some linters, as we discussed in https://github.com/astronomer/dag-factory/pull/433

docs/features/cli.md
```diff
@@ -2,18 +2,62 @@
 
 After installing DAG Factory, the CLI can be invoked using the `dagfactory` command.
 
-## Usage
+## Commands summary
+
+| Command | Args   | Flags       | Description                                          |
+| ------- | ------ | ----------- | ---------------------------------------------------- |
+| `lint`  | `path` | `--verbose` | Check if the given directory or file is a valid YAML |
```
pankajastro, line 9: Not in this PR, but I feel it may be worth providing an option to ignore some files

pankajastro (approved):

docs/features/cli.md
tatiana: @pankajastro I was worried about having -v with two meanings: `--version` and `--verbose`. I believe it is more useful when using `--verbose`, since it seems a more common use case - what are your thoughts?

dagfactory/_yaml.py
tatiana: Now is the time for us to review this strategy!

dagfactory/_yaml.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
+
+import os
+
+import yaml
+
+from .utils import cast_with_type
+
+
+def load_yaml_file(file_path: str) -> dict[str, any]:
```
pankajkoti, line 10: probably could also make this method private

pankajkoti (approved):
LGTM, minor suggestion inline about considering making the method private.

dagfactory/_yaml.py
tatiana: Addressed in 9caa6b8fddf2db8c397b1053d6db3f5018a5728d

docs/features/cli.md
tatiana: Thanks, @pankajastro , I added a follow up ticket: https://github.com/astronomer/dag-factory/issues/516

dagfactory/_yaml.py
tatiana: I logged this ticket so we can further discuss:
https://github.com/astronomer/dag-factory/issues/517

PR #522, opened 2025-07-30
Add missing env in contributing doc

pankajastro:


docs/contributing/howto.md
```diff
@@ -141,6 +141,14 @@ The [pyproject. toml](https://github.com/astronomer/dag-factory/blob/main/pyproj
 
 ### Run unit tests
 
+> Note: these tests create local Python virtual environments in a hatch-managed directory.
+
+First, set the following environment variables:
+
+```bash
+export AUTO_CONVERT_TO_AF3=true
```
pankajkoti, line 150: Do we need to always set this for running unit tests or only for Airflow 3 based hatch envs?

docs/contributing/howto.md
```diff
@@ -161,9 +169,11 @@ hatch run tests:test-cov
 First, set the following environment variables:
 
 ```bash
+export AIRFLOW__CORE__DAGBAG_IMPORT_TIMEOUT=90
 export AIRFLOW_HOME=$(pwd)/dev/
-export CONFIG_ROOT_DIR=`pwd`"/dev/dags"
-export PYTHONPATH=dev/dags:$PYTHONPATH
+export CONFIG_ROOT_DIR=$(pwd)/dev/dags
+export PYTHONPATH=$(pwd)/dev/dags:$PYTHONPATH
+export AUTO_CONVERT_TO_AF3=true
```
pankajkoti, line 179: same question as above

docs/contributing/howto.md
pankajastro: For Airflow 3

pankajkoti: Then would be nice to specifically mention that in the docs. Otherwise, it leaves an impression that we always need to set that. WDYT?

pankajastro: Added in note section

mkdocs.yml
```diff
@@ -33,6 +33,8 @@ extra:
     alias: true
 
 markdown_extensions:
+  - admonition
+  - codehilite
```
pankajkoti, line 37: What are these extensions for? Can we add inline comments for their purpose please if possible?

docs/contributing/howto.md
```diff
@@ -141,6 +141,16 @@ The [pyproject. toml](https://github.com/astronomer/dag-factory/blob/main/pyproj
 
 ### Run unit tests
 
+!!! note
+    - These tests create local Python virtual environments in a hatch-managed directory.
+    - The `AUTO_CONVERT_TO_AF3` environment variable is required to run tests in the Airflow 3 environment.
+
+First, set the following environment variables:
```
pankajkoti:
```suggestion

If you have YAMLs written for Airflow 2 and would like them to be run for Airflow 3 tests, set the following environment variable for DAG Factory to convert and make them compatible with Airflow 3:
```

mkdocs.yml
pankajastro: To render notes better way.

pankajastro: Example
- https://astronomer.github.io/dag-factory/dev/contributing/howto/#2-activate-the-local-python-environment
vs
- https://squidfunk.github.io/mkdocs-material/reference/admonitions/

PR #523, opened 2025-07-30
Refactor KPO to use __type__ syntax

pankajastro:
- This PR refactors the KPO implementation to leverage the `__type__` syntax
- Clean the custom type conversion logic
- It removes multiple try/except blocks in the DAG-Factory codebase
- This change also helps offload compatibility for airflow-providers-cncf-kubernetes from the DAG-Factory codebase
- Add example docs for KPO


dev/dags/kpo.yml
```diff
@@ -0,0 +1,23 @@
+kubernetes_pod_dag:
+  start_date: 2025-01-01
+  schedule_interval: "@daily"
+  description: "A DAG that runs a simple KubernetesPodOperator task"
+  catchup: false
+  tasks:
+    hello-world-pod:
+      operator: airflow.providers.cncf.kubernetes.operators.pod.KubernetesPodOperator
+      config_file: "path/to/kube/config"
```
pankajkoti: Can we please have the tasks as list notation here?

pankajkoti (approved):
Looks solid and much cleaner! Looking forward to users trying it out and leaving their feedback!

dev/dags/kpo.yml
pankajastro: changed to list: 946f86e64b4d4091e0834202982eb40fd96d7491

PR #524, opened 2025-07-31
Supporting extended configuration files

### LINKED ISSUE #443: [Feature] Allow multiple shared defaults YAML
### Description

Having all shared defaults in one file makes it challenging for users who manage multiple teams or projects with dag-factory. Also, https://github.com/astronomer/dag-factory/issues/441 proposed to support both `defaults.yml` and `defaults.yaml`, while it's not decided yet how to handle multiple defaults files.

This feature request proposes a new way handling defaults, which supports multiple default files, allowing custom defaults to be used, and also allowing using default in a DAG YAML as the default of others. Specifically, a new reserved field `__extends__` is added accepting multiple default config. The extended default config can have any custom file name.

```
# dag_1.yaml

__extends__:
    - default_1.yaml
    - project1/project_default.yaml

default:
  default_args:
    start_date: '2024-01-01'
```

Chained config extending can be allowed:
```
# default_1.yaml

__extends__:
    - base_default.yaml
```

Further more, the extended config can not only be default config with only `default` specified, but can also be an actual DAG config file, because only `default` will be applied during extending, and other fields will be ignored in the downstream config.

For the backward compatibility, technically, existing `defaults.yml` can still be supported. While it can also be confusing supporting both ways. I think it's discussable whether we want to keep a global default config. Personally, I think having both can cover more use cases and preferences, while we need to figure out how to support defaults with both yml and yaml suffix.

### Use case/motivation

For users who use dag-factory for multiple teams/projects, they have the need to setup different shared defaults per teams or projects. This is not possible with current defaults.yml solution, as it can only be one file.

### Related issues

https://github.com/astronomer/dag-factory/issues/441
https://github.com/astronomer/dag-factory/issues/297

gyli:
Resolves https://github.com/astronomer/dag-factory/issues/443

Note that in this PR, `__extends__` only supports path relative to the YAML file directory. In the future, we can add a new argument to `DagFactory`, accepting the base dir for extends.


dagfactory/utils.py
```diff
@@ -426,3 +427,21 @@ def cast_with_type(data):
         return [cast_with_type(item) for item in data]
 
     return data
+
+
+def merge_dict(dict1: dict, dict2: dict) -> dict:
```
tatiana: I believe we should avoid extending the definitions in the `utils.py` module. My vote is usually not not have this type of module. It tends to grow, and it is hard to track its contents.

I believe we should rename this method ot `_merge_dicts`, define ti inside the `DagFactory` class and join the implementation done in 
https://github.com/astronomer/dag-factory/blob/2fd9c757d564eed1686f24edf353d166db518d64/dagfactory/dagfactory.py#L138-L163

With this one.

The advantage of this one is that it handles generically merges of values that map to lists and dictionaries (the other two functions were specific to certain keys).

The advantage of the other `_merge` methods is that they handle a list of dictionaries. We should combine both to get the best combination.

dagfactory/utils.py
```diff
@@ -426,3 +427,21 @@ def cast_with_type(data):
...
+    """
+    Recursively merges dict2 into dict1.
+    It does not update dict1 in place, but returns a new dict.
+    """
+    result = copy.deepcopy(dict1)
+
+    def recursive_dict_merge(_dict1: dict, _dict2: dict) -> None:
```
tatiana: It would be great if we could avoid recursion. Historically, recursion is not performant in Python (an example is discussed [in this blog post](https://neopythonic.blogspot.com/2009/04/tail-recursion-elimination.html)).

Do we really need to expose this in DAG-level args, including `default_args`? What are the use-cases?

dagfactory/dagfactory.py
```diff
@@ -207,6 +208,18 @@ def _load_dag_config(self, config_filepath: str) -> Dict[str, Any]:
...
+            extend_config_queue = copy.deepcopy(config.get(EXTENDS_KEY, []))
+
+            while extend_config_queue:
+                extend_config_relative_path = extend_config_queue.pop(0)
+                extend_config_path = os.path.join(os.path.dirname(config_filepath), extend_config_relative_path)
+                extend_config_path = os.path.abspath(extend_config_path)  # Ensure absolute path
+                extend_config = DagFactory(config_filepath=extend_config_path).config
+                # Only 'default' is recognized in the extended config
+                config["default"] = merge_dict(extend_config.get("default", {}), config.get("default", {}))
+                extend_config_queue.extend(extend_config.get(EXTENDS_KEY, []))
```
tatiana: I believe we should make the code consistent and these lines should not be here.

We can isolate the logic that extracts the yaml file names defined in  `__extends__` into a method similar to `_retrieve_default_yaml_filepaths`. We could perhaps:
* Rename the existing `_retrieve_default_yaml_filepaths` to  `_retrieve_default_yaml_filepaths_from_directory_hierarchy`
* Add a new `_retrieve_default_yaml_filepaths_from_extends`

I believe the remaining of the logic, that decides if we're using `EXTENDS_KEY` and invokes the auxiliary merge method should be defined here:
```
https://github.com/astronomer/dag-factory/blob/2fd9c757d564eed1686f24edf353d166db518d64/dagfactory/dagfactory.py#L76-L89
```

tatiana (commented):
@gyli Thanks a lot for the work on this, not only including the implementation, but also tests and examples.

I wrote feedback on the implementation, particularly on parts that we could improve so both #500 and this PR would reuse most of the code, and it would be easier to maintain. It is worth having a separate PR, for instance, on refactoring the merge methods - since this is a significant contribution that will be accepted without further discussion.

DagFactory currently has four different ways of defining defaults:
https://astronomer.github.io/dag-factory/dev/configuration/defaults/

And, I agree with @pankajkoti, to add another one will make things even more complex. We've been trying to reduce entry points in the 1.0 release (such as in #509), and keeping both `defaults.yml` implementations may be overwhelming for users & will add some maintenance burden (which I believe is relative small, if we refactor the code).

I do not have a preference between the interfaces proposed in #469 and #443. Ideally, we'd validate these with end-users and make a user-driven / data-driven decision. Unfortunately, we don't have time to do this before the 1.0 release. I agree with the pros and cons you and @pankajkoti raised, including that #443 is more extensible, but more complex to use.

We had discussed this feature with our new product manager, @yetudada, earlier this week, and due to the lack of data, we had deferred #443 to the 1.1 release, three days ago (as seen in the ticket #443).

I can see two main paths forward:
1. Keep one of the interfaces (#443 or #469) for the 1.0 release, measure adoption, and consider introducing the other one afterwards (e.g. 1.1 release). In this case we need to choose one of them.
2. Release both features  (#443 or #469) as experimental and have a way of measuring adoption, and remove one of them after collecting data.

@yetudada @pankajastro @pankajkoti - I know we discussed these three days ago, but by then we didn't have a PR. Circumstances changed, and I'd love to know if you see any other path forward. Ultimately, this is a product decision.

docs/configuration/defaults.md
```diff
@@ -154,3 +156,70 @@ Given the various ways to specify top-level DAG arguments, including `default_ar
 2. In the `default` block within the workflow's YAML file
 3. The arguments defined in `default_args_config_dict`
 4. If (3) is not declared, the `defaults.yml` hierarchy.
+
+
+## Configuration Inheritance with `__extends__`
```
tatiana, line 161: I believe this documentation would have to live in "### Combining multiple `defaults.yml` files", and we'd either replace the current description or we'd have both as sub-cases of this broader approach.

pankajastro:
Both approaches aim to reduce duplication and improve reusability.

However, keeping both `__extend__` and default feels a bit redundant. It also increases the number of possible permutations, which could lead to confusion. It would be cleaner if we could stick to a single concept—either `__extend__` or `default`.

Personally, I lean toward something more explicit and controllable, but I might be biased since I haven’t authored many YAML files myself. The `__extend__` approach also reminds me of [Docker Compose's include mechanism](https://docs.docker.com/compose/how-tos/multiple-compose-files/include/), which feels familiar.

@gyli — are we planning to deprecate defaults.yml, or how do we plan to handle it?

If we go with the `__extend__` approach, it might introduce one more breaking change that we’ll need to account for.

@yetudada, looking forward to hearing your thoughts.

gyli:
I think these two options make sense to us:
1. Decide our preferred one in this thread before merging. (Can we collect feedback from any known users?)
2. Merge this and keep both approaches for the short term (a few months?), maybe until next version. (Do we have any timeline planned?)

Meanwhile, I will update this PR based on @tatiana 's comment to reuse more existing utils.

dagfactory/utils.py
gyli: This is a recently requested feature: https://github.com/astronomer/dag-factory/issues/462. 
PR #500 rolled back this feature. No matter it's intentional or not, this rollback should be noted down in the doc or changelog to avoid the confusion.
cc @pankajastro

dagfactory/dagfactory.py
gyli: I will move the logic out of `_load_dag_config`, while I prefer not to separate the logic into retrieving file paths and loading extended files, because it will require loading every extended files twice.

dagfactory/utils.py
gyli: code updated

dagfactory/utils.py
gyli: code updated and recursion removed. This is possible because there's no nested dict value in current dag-level args.

docs/configuration/defaults.md
gyli: Having both as sub-cases is exactly what I tried to do here. It has these H2 level title:
```
## YAML top-level `default`
## Specifying `default` arguments via a Python dictionary
## Declaring default values using the `defaults.yml` file
## Configuration Inheritance with `__extends__`
```

PR #525, opened 2025-07-31
Remove custom yml constructor and consolidate boolean implementation

### LINKED ISSUE #507: Consolidate syntax for logical operator
### Description

PR https://github.com/astronomer/dag-factory/pull/413 added logical operator `and`, `or` for asset, while data-aware scheduling uses different logical operator `!and`, `!or`, and `!join`.

Although they have different logic behind the operators (one init Asset objects in a list, while the other converts `!and` to `&`), I think we should keep a unified syntax for logical operators with all Airflow modules. One solution is to replace `!and`, `!or` with `and`, `or`.

### Use case/motivation

_No response_

### Related issues

_No response_

- [ ] Yes, I am willing to submit a PR!

### LINKED ISSUE #517: Review the boolean logic approach
Based on this discussion:
- https://github.com/astronomer/dag-factory/pull/513#discussion_r2239044811

DAG Factory YAML may not be parseable using all tools, particularly due to our boolean logic.

Do we want to change this for 1.0?

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/517
closes: https://github.com/astronomer/dag-factory/issues/507

This PR removes the use of the custom YAML parser for the tags `!and`, `!or`, and `!join`.

Introduced dag-factory native parsing to parse `__and__`, `__or__`, and `__join__` syntax for parsing DAG configuration.

This makes DAG YAML linting compatible with all linters.

### DIFF
--- a/dagfactory/_yaml.py
+++ b/dagfactory/_yaml.py
@@ -23,16 +23,16 @@ def load_yaml_file(file_path: str) -> dict[str, any]:
                 try:
                     return reduce(lambda a, b: a & b, processed_items)
                 except TypeError:
-                    return f"({' & '.join(processed_items)})"
+                    return f"({' & '.join(str(item) for item in processed_items)})"
             if "__or__" in data:
                 processed_items = [_flatten_logical_expressions_helper(item) for item in data["__or__"]]
                 try:
                     return reduce(lambda a, b: a | b, processed_items)
                 except TypeError:
-                    return f"({' | '.join(processed_items)})"
+                    return f"({' | '.join(str(item) for item in processed_items)})"
             if "__join__" in data:
                 processed_items = [_flatten_logical_expressions_helper(item) for item in data["__join__"]]
-                return "".join(processed_items)
+                return "".join(str(item) for item in processed_items)
             new_dict = {}
             for key, value in data.items():
                 new_dict[key] = _flatten_logical_expressions_helper(value)
--- a/dagfactory/_yaml.py
+++ b/dagfactory/_yaml.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import copy
 import os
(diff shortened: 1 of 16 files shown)


tatiana (approved):
Thanks, @pankajastro , looks good!

This change is consistent with DagFactory YAMLs being lintable in any tool. Similar to when we decided to use implementation #444 instead of #443 for type casting.

It's a shame that most linters (including PyYAML) are sensitive to custom constructs that are YAML-Spec compatible, and error if they do not know how to handle them.

 It seems a simple migration from the old implementation from the new one.

It would be great to get any thoughts from @ErickSeo , who originally implemented this feature in #320.

pankajastro:
> About migration, I believe that implementing will also be quite simple.
>
>The question that remains is whether it’s worth providing a fallback for the previous solution. At a glance, I think that point >was missing—or I may have overlooked something
>
>It really comes down to the product versus user perspective.
>
>To make things easier for the user, the lib could either keep things the same (or show a warning that says that, starting >from a certain version, YAML files using the !and, !or, and !join operators won't be supported anymore).


We want to keep the migration process as simple as possible for users, while also reducing the maintenance burden on the maintainers.

Yes, the fallback is not provided for two main reasons:

- We aim to reduce the long-term maintenance overhead.
- The upcoming release is a major version. If we don't remove `!or`, `!and`, and `!join` now, we’ll have to wait for the next major release, as it's not ideal to remove them in a minor version.

@tatiana - please correct me if I'm wrong, but since we're providing some utilities via the CLI to improve developer experience, I think it might be helpful to expose something in the CLI to support this migration. That said, I'm not sure if we currently have bandwidth for it.

I'm happy to support the community with reviewing, testing, if anyone wants to contribute.

PR #526, opened 2025-07-31
Bugfix - Dataset Names with Dots Caused Orphan Nodes in AST

ErickSeo:
### Bugfix - Dataset Names with Dots Caused Orphan Nodes in AST  

A bug was found where dataset names containing a dot (.) (e.g., `main.data.user_logs`) were not properly recognized by the regex used in our dataset extraction utilities.  

The old regex would split the dataset name at the dot, so only the first part (e.g., `main`) was extracted. 
The remaining segments (`data.user_logs`) were left as separate nodes, resulting in "orphan nodes" when parsing dataset expressions into an AST (e.g., when using SafeEvalVisitor).  
This caused parsing errors and broke support for hierarchical or namespaced dataset names.

### What was fixed

- Updated the regex in `extract_dataset_names` to correctly match full dataset names that include dots, dashes, slashes, and similar characters.

- Now, hierarchical names like `main.data.user_logs` are recognized as a single dataset, preventing orphan nodes during AST parsing.

#### Before:

- Input: `"main.data.user_logs"`
- Extracted: `['main', 'data', 'user_logs']`
- Result: The AST parser split the dataset name into multiple orphan nodes (`main`, `data`, `user_logs`), instead of a single node.
- Impact: Any downstream logic or dependencies expecting the full dataset name as `"main.data.user_logs"` would fail to find it, since the name was split into separate nodes and the expected full name node did not exist.

#### After:

- Input: "main.data.user_logs"
- Extracted: ['main.data.user_logs']
- Result: The full name is parsed as a single node; no orphans.

### Summary
This PR fixes dataset extraction and AST parsing issues for hierarchical dataset names. Now, expressions and tools like SafeEvalVisitor can handle complex dataset names reliably.


ErickSeo:
I am working on a refactor to improve code readability and make the dataset logic (triggers/schedules, inlets, outlets) easier to understand and maintain. 
This PR is part of that effort and fixes a bug where hierarchical dataset names were being incorrectly split into separate nodes in the AST.

I will open an issue soon to document this bug and link this PR and the related changes, so the context and fixes are fully documented.

pankajastro:
What do you think about leveraging the `__type__` syntax for the datasets? Custom conversions might introduce bugs and will be difficult to maintain. Additionally, the dag-factory would always need to handle any future compatibility issues. I believe using the `__type__` syntax would reduce maintenance costs for this repo, and the interface will be aligned more closely with Airflow docs

```yaml
schedule:
    __type__: airflow.datasets. Dataset
    uri: s3://
    name: ...
```

ErickSeo:
> What do you think about leveraging the `__type__` syntax for the datasets? Custom conversions might introduce bugs and will be difficult to maintain. Additionally, the dag-factory would always need to handle any future compatibility issues. I believe using the `__type__` syntax would reduce maintenance costs for this repo, and the interface will be aligned more closely with Airflow docs
> 
> ```yaml
> schedule:
>     __type__: airflow.datasets. Dataset
>     uri: s3://
>     name: ...
> ```

I’m already exploring the __type__ approach in a separate branch. Over there I’m refactoring things so the legacy logic remains compatible with your 0.23 implementation. 

```python
 @staticmethod
    def _asset_schedule(value):
        """Recursively parse and construct assets or combinations of assets."""
        from airflow.sdk import Asset

        if isinstance(value, dict):
            if "or" in value:
                assets = [DagBuilder._asset_schedule(item) for item in value["or"]]
                return DagBuilder._combine_assets(assets, "or")
            elif "and" in value:
                assets = [DagBuilder._asset_schedule(item) for item in value["and"]]
                return DagBuilder._combine_assets(assets, "and")
        elif isinstance(value, list):
            return [asset for asset in value]
        elif isinstance(value, Asset):
            return value
        else:
            raise TypeError(f"Unexpected data type: {type(value)}")
```

I’ve also started pulling the schedule functionality into its own domain modules to make maintenance easier. 
I’ll have a preview ready for you later this week.

For clarity, I split the refactor and the bug-fix into two separate branches so we don’t mix up what’s purely a refactoring effort versus what’s a targeted fix.

tatiana (commented):
@ErickSeo thanks for the detailed PR description. I'm aware this PR is in draft. When you have a chance, please, add an example of a YAML that broke to the PR description or as a test.

PR #532, opened 2025-08-01
Add migration guide docs

pankajastro:
Helper docs for migrating to DAG-Factory 1.0.

This includes the example and other details that can help users.


tatiana (approved):
This looks great, @pankajastro , thank you!

It would be great if we extended #539 to have an option of converting between dagfactory0to1 YAML files as well. It would not cover all the changes described in this document, but could automatically do conversions like `execution_delta_secs`. If you think it is a good idea, we could log a separate ticket to cover this.

PR #533, opened 2025-08-01
Remove custom timetable parsing

pankajastro:
This PR remove the custom parsing for the Airflow timetable object.
The previous parsing supports kwargs only, and was also very unpredictable
These changes make the use of the timetable object consistent with `__type__`
This issue was reported in [doc](https://docs.google.com/document/d/1u_9Xn0U6y4hEz3GWmPA4D7c73U1yN_f89ACXHAxrq7w/edit?tab=t.0)

Airflow 2
```yaml
default:
  default_args:
    start_date: 2024-11-11

basic_example_dag:
  timetable:
    __type__: airflow.timetables.trigger.CronTriggerTimetable
    __args__:
      - 0 1 * * 3
    timezone: UTC
  catchup: false
  tasks:
    - task_id: "task_1"
      bash_command: "echo 1"
```

Airflow 3

```yaml
default:
  default_args:
    start_date: 2024-11-11

basic_example_dag:
  schedule:
    __type__: airflow.timetables.trigger.CronTriggerTimetable
    __args__:
      - 0 1 * * 3
    timezone: UTC
  catchup: false
  tasks:
    - task_id: "task_1"
      operator: airflow.providers.standard.operators.bash.BashOperator
      bash_command: "echo 1"
```


tatiana (commented):
@pankajastro thanks a lot for addressing this.

Please, could you rebase? I believe the Airflow 3 example in the description will need to change - given that the import path seems to be specific to Airflow 2 and we are no longer automatically converting between the two.

Also, would it make sense to add a test-case or example DAG to cover the current timetable conversion?

pankajastro:
> @pankajastro thanks a lot for addressing this.
> 
> Please, could you rebase? I believe the Airflow 3 example in the description will need to change - given that the import path seems to be specific to Airflow 2 and we are no longer automatically converting between the two.
> 
> Also, would it make sense to add a test-case or example DAG to cover the current timetable conversion?

Right, changed the path in description and added example DAG

tatiana (approved):
Looks great, @pankajastro , thank you very much!

Should we add this change to the migration guide as well?
https://astronomer.github.io/dag-factory/dev/migration_guide/
Happy for this to be done on a separate PR.

pankajastro:
> Should we add this change to the migration guide as well? https://astronomer.github.io/dag-factory/dev/migration_guide/ Happy for this to be done on a separate PR.

Created a PR: https://github.com/astronomer/dag-factory/pull/549

PR #538, opened 2025-08-02
Document the Asset example DAG

pankajastro:
This PR documents the Asset example DAG

Screenshot 2025-08-03 at 3 23 30 AM


pankajkoti (approved):
LGTM. Please check if https://github.com/astronomer/dag-factory/pull/509 might have an effect on the way we show on generating the DAG in the documentation

PR #539, opened 2025-08-04
Add `convert` CLI command to migrate from af2 to af3

### LINKED ISSUE #445: [Feature] DAG-factory CLI

tatiana:
Introduce the `convert` command with the optional `--override` flag, that would apply the changes in-place.

Similar to Python [2to3](https://docs.python.org/3.12/library/2to3.html), the files are not overridden by default.

Examples of usage:

1. Show differences in the command line without converting in-place:
Screenshot 2025-08-04 at 09 58 44

2. Identify that no changes are needed:
Screenshot 2025-08-04 at 09 59 06

3. Exit with error code 1 and describe the error if the given YAML is found:
Screenshot 2025-08-04 at 09 59 38

4. Change multiple files within a directory:
Screenshot 2025-08-04 at 10 00 01

5. Change files in place
Screenshot 2025-08-04 at 10 00 24

As agreed, the documentation will come in a separate PR.

Closes: #445


dagfactory/__main__.py
```diff
@@ -108,5 +111,58 @@ def lint(
...
+                        if line.startswith("+") and not line.startswith("+++"):
+                            console.print(Text(line, style="green"))
+                        elif line.startswith("-") and not line.startswith("---"):
+                            console.print(Text(line, style="red"))
+                        else:
+                            console.print(line)
+                else:
+                    console.print(f"[blue]No changes needed:[/blue] {file}")
+
+        except Exception as e:
```
pankajastro, line 172: Would it be possible to make `Exception` more specific?

dagfactory/__main__.py
```diff
@@ -108,5 +111,58 @@ def lint(
         console.print(f"Analysed {len(files)} files, [green]no errors found.[/green]")
 
 
+@app.command()
+def convert(
+    path: Path = typer.Argument(..., help="Path to a YAML file or a directory of YAML files to convert"),
+    # type: str = typer.Option("airflow2to3", "--type", "-t", help="Conversion type (default: airflow2to3)"),
```
pankajastro, line 127: This PR only support `airflow2to3`? Would it make sense to update the docstring accordingly?

dagfactory/__main__.py
```diff
@@ -108,5 +111,58 @@ def lint(
...
+):
+    """Convert YAML files using a specific conversion type."""
+    files = _find_yaml_files(path)
+    failures = 0
+
+    for file in files:
+        try:
+            original_data = load_yaml_file(file)
+            # we need to create a copy because the `update_yaml_structure` modifies the content by reference
+            converted_data = update_yaml_structure(deepcopy(original_data))
```
pankajastro, line 139: Method `update_yaml_structure` was created for our CI only. Do we need to check if it works for objects like asset, datasets, etc?

tatiana: @pankajastro, considering our YAML `dev/dags` files were representative enough, these should be covered. If we were using to convert things to the CI, it already shows a good coverage. I believe it is outside of the scope of this ticket to improve the conversion logic further. This can be improved incrementally, based on users' feedback.

dagfactory/__main__.py
tatiana: I fully agree with not using `Exception` in most backend systems and libraries. However, in a CLI, I believe the experience can be quite bad if we don't capture exceptions - stack traces can be overwhelming. For this reason, I left this particular exception as it is. Given that we output the exception string straight after, users would be able to identify the cause, so we would not be hiding information from the end-user, just giving a better experience.

pankajastro (approved):

dagfactory/__main__.py
tatiana: I added a follow up ticket: https://github.com/astronomer/dag-factory/issues/542

PR #541, opened 2025-08-04
Add docs for the CLI convert command

tatiana:
Introduced in #539

### DIFF
--- a/docs/features/cli.md
+++ b/docs/features/cli.md
@@ -4,11 +4,63 @@ After installing DAG Factory, the CLI can be invoked using the `dagfactory` comm
 
 ## Commands summary
 
-| Command | Args   | Flags       | Description                                          |
-| ------- | ------ | ----------- | ---------------------------------------------------- |
-| `lint`  | `path` | `--verbose` | Check if the given directory or file is a valid YAML |
+| Command    | Args   | Flags        | Description                                                          |
+| ---------- | ------ | ------------ | -------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


docs/features/cli.md
```diff
@@ -57,7 +24,7 @@ dagfactory [OPTIONS]
 | `--version` |       | Show the installed version of DAG Factory and exit |
 | `--help`    | `-h`  | Show this message and exit                         |
 
-### Version
+#### Identify the CLI version
```
pankajkoti, line 27: Is it the CLI's version or the dag-factory version? I guess we won't maintain CLI versions?

pankajkoti (approved):
Nice one, thanks for the docs! 👏🏽 

One quick question in-line.

docs/features/cli.md
tatiana: Both the DAG Factory and the CLI versions are the same - similar to Apache Airflow and dbt Core.

pankajkoti: yes, happy for it to be as is now. We can address it later if it confuses someone.

docs/features/cli.md
```diff
@@ -57,7 +24,7 @@ dagfactory [OPTIONS]
...
 
 Display the DAG Factory version:
```
tatiana:
```suggestion
Display the DAG Factory version (both the CLI and the library share the same version number):
```

docs/features/cli.md
tatiana: @pankajkoti I just improved the description in 9a6de0c427f65575c8d37e2b39cee174f4793bff, thanks for the feedback!

PR #546, opened 2025-08-05
Rename parameter in `load_yaml_dags` to reflect behaviour

### LINKED ISSUE #505: Rename `default_args_config_path` and `default_args_config_dict`
During the development of #500, I noticed that  `default_args_config_path` and `default_args_config_dict` support overriding not only `default_args`, but also any DAG-level args.

This is a proposal, but I'm open-minded:
- `defaults_config_path`
- `defaults_config_dict`

Also, we should do this as part of #472.

Since this is potentially a breaking change, we should consider implementing it as part of Milestone 1.0.

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/505

Changes `_DagFactory` class param
- `config` -> `config_dict`
- `default_args_config_path` -> `defaults_config_path`
- `default_args_config_dict ` -> `defaults_config_dict `

Changes `load_yaml_dags` param
- `default_args_config_dict` -> `default_config_dict`

### DIFF
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -30,46 +30,46 @@ class _DagFactory:
     :param config_filepath: the filepath of the DAG factory YAML config file.
         Must be absolute path to file. Cannot be used with `config`.
     :type config_filepath: str
-    :param config: DAG factory config dictionary. Cannot be used with `config_filepath`.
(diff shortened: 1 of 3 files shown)


tatiana (approved):
Looks good, thanks a lot, @pankajastro !

Issue #550, opened 2025-08-08
[Bug] Pendulum 3.1.0 compatibility issue with Airflow 2.4.3


viiccwen:
### DAG Factory version

0.23.0

### airflow version

2.4

### Python version

3.11

### Deployment

Virtualenv installation

### Deployment details

follow the [Contributing Guide](https://astronomer.github.io/dag-factory/latest/contributing/howto/#using-a-python-virtual-environment-for-local-development) to build the environment with `uv` virtual environment.

### What happened?

When running tests with Airflow 2.4.3, we encounter a `TypeError: 'module' object is not callable` error in Airflow's `settings.py` file at line 49:

```python
TIMEZONE = pendulum.tz.timezone('UTC')
```

this issue only happens in Python 3.11 with AirFlow 2.4, I found this project use **pendulum 3.1.0** with AirFlow 2.4 version.

> It's the issue of AirFlow 2.4, see [cannot start 'airflow standalone'; pendulum.tz.timezone(tz) -> TypeError](https://github.com/apache/airflow/issues/38714)

### Error Details

```
TypeError: 'module' object is not callable
File: airflow/settings.py:49
Code: TIMEZONE = pendulum.tz.timezone('UTC')
```

### Relevant log output

```shell
_________________________ ERROR collecting dev/tests/dags/test_dag_example.py __________________________
dev/tests/dags/test_dag_example.py:8: in <module>
    from airflow.models import DagBag
../../../Library/Application Support/hatch/env/virtual/dag-factory/nMdJ8nG1/tests.py3.11-2.4/lib/python3.11/site-packages/airflow/__init__.py:34: in <module>
    from airflow import settings
../../../Library/Application Support/hatch/env/virtual/dag-factory/nMdJ8nG1/tests.py3.11-2.4/lib/python3.11/site-packages/airflow/settings.py:49: in <module>
    TIMEZONE = pendulum.tz.timezone('UTC')
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   TypeError: 'module' object is not callable
________________________________ ERROR collecting tests/test__main__.py ________________________________
tests/test__main__.py:10: in <module>
    from dagfactory import __version__
dagfactory/__init__.py:3: in <module>
    from .dagfactory import load_yaml_dags
dagfactory/dagfactory.py:13: in <module>
    from airflow.sdk.definitions.dag import DAG
[...]

PR #551, opened 2025-08-08
Fix Pendulum compatibility issue for Airflow 2.4

viiccwen:
### Description

This PR solves a compatibility issue between **Airflow 2.4** and **Pendulum**. Specifically, Airflow 2.4 is not fully compatible with Pendulum version 3.0.0 and above.

### Changes:

* Modified `/tmp/constraint.txt` to remove `pendulum==` version restriction.
* Added command to upgrade `pip` using `uv`.
* Installed compatible Pendulum version `<3.0.0` if Airflow 2.4 is detected.

### Why:

* Airflow 2.4 has compatibility issues with Pendulum version 3.0.0 and above. This fix ensures that the correct version of Pendulum is installed and works seamlessly with Airflow 2.4.

### Related Issues:

#550


pankajastro (commented):
Thanks for the PR, @viiccwen! I'm curious why the CI isn't detecting it -- could it be because you're using Python 3.11?

Issue #554, opened 2025-08-10
[Bug] Incorrect usage of `load_yaml_dags` parameters causing integration test to miss DAGs


viiccwen:
### DAG Factory version

0.23.0

### airflow version

2.10.5

### Python version

3.12.7

### Deployment

Virtualenv installation

### Deployment details

follow the [contribute guide](https://astronomer.github.io/dag-factory/latest/contributing/howto/#overview-of-how-to-contribute) to build local development environment with uv & hatch.

### What happened?

The `load_yaml_dags` function is being passed an incorrect parameter in `dev/dags/example_object_storage.py`, `dev/dags/exampe_timetable_schedule.py`. A single file path was provided for the `dags_folder` parameter, which is intended to be used for folder paths. This prevented the function from recursively scanning for YAML files, leading to the DAG not being loaded and, as a result, not being tested in the integration tests.

#### Impact

* The related DAG was not included in the integration tests, causing it to be skipped and not tested.
* This issue prevents the accurate testing of DAG functionality and might lead to unnoticed issues in the pipeline.

### How to reproduce

1. follow the [Contributing Guide](https://astronomer.github.io/dag-factory/latest/contributing/howto/#using-a-python-virtual-environment-for-local-development) to build the environment with uv virtual environment.
2. run hatch run tests.py3.9-3.0:test-cov in CLI

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

PR #555, opened 2025-08-10
Fix incorrect `load_yaml_dags` parameter usage

viiccwen:
### Description
The `load_yaml_dags` function in `dev/dags/example_object_storage.py` was incorrectly provided with a single file path for the `dags_folder` parameter. This parameter is intended for folder paths and was causing the function to skip the recursive scanning of YAML files. As a result, the associated DAG was not being loaded and therefore not included in the integration tests.

### Changes
change from `dags_folder` to `config_filepath` parameter to ensure `example_object_storage` could be included in integration testing.

### Related Issue
Closes: #554

### DIFF
--- a/tests/test_example_dags.py
+++ b/tests/test_example_dags.py
@@ -30,6 +30,10 @@ MIN_VER_DAG_FILE_VER: dict[str, list[str]] = {
     "2.9": ["example_map_index_template.py", "example_object_storage.py"],
 }
 
+MAX_VER_DAG_FILE_VER: dict[str, list[str]] = {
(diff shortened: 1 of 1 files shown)


viiccwen:
for short term, right? That's a good idea : )
I'll remove 
[ignored exampe_timetable_schedule testing in AirFlow 3.0](https://github.com/astronomer/dag-factory/pull/555/commits/97f441f3008e7624df7f5b03b0834318e6b2272b) 
after changing the location of `example_timetable_schedule.yml`

viiccwen:
Hello @pankajastro,
Any updates or anything I could help?

PR #567, opened 2025-08-25
Allow yaml/yml suffix for shared defaults

### LINKED ISSUE #441: [Feature] Allow yaml/yml suffix for shared defaults
### Description

File name defaults.yml is hard coded, with only one file extension name supported. While both `.yaml` and `.yml` could be YAML file and should be supported.

### Use case/motivation

_No response_

### Related issues

https://github.com/astronomer/dag-factory/pull/330/files

pankajastro:
closes: https://github.com/astronomer/dag-factory/issues/441

This PR enable support for the suffix `yaml` and `yml` for the default file. Before this PR, only `yml` was supported. The `yml` extension will be preferred over `yaml`

### DIFF
--- a/dagfactory/constants.py
+++ b/dagfactory/constants.py
@@ -4,4 +4,4 @@ TELEMETRY_TIMEOUT = 1.0
 
 AIRFLOW3_MAJOR_VERSION = 3
 
-DEFAULTS_FILE_NAMES = ["defaults.yml", "defaults"]
+DEFAULTS_FILE_NAMES = ["defaults.yml", "defaults.yaml"]
--- a/dagfactory/constants.py
+++ b/dagfactory/constants.py
@@ -4,4 +4,4 @@ TELEMETRY_TIMEOUT = 1.0
 
 AIRFLOW3_MAJOR_VERSION = 3
 
-DEFAULTS_FILE_NAME = "defaults.yml"
+DEFAULTS_FILE_NAMES = ["defaults.yml", "defaults"]
--- a/dagfactory/dagfactory.py
+++ b/dagfactory/dagfactory.py
@@ -15,7 +15,7 @@ except ImportError:
     from airflow.models import DAG
 
 from dagfactory._yaml import load_yaml_file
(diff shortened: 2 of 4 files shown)


dagfactory/constants.py
```diff
@@ -4,4 +4,4 @@
 
 AIRFLOW3_MAJOR_VERSION = 3
 
-DEFAULTS_FILE_NAME = "defaults.yml"
+DEFAULTS_FILE_NAMES = ["defaults.yml", "defaults"]
```
pankajastro:
```suggestion
DEFAULTS_FILE_NAMES = ["defaults.yml", "defaults.yaml"]
```

dagfactory/dagfactory.py
```diff
@@ -110,14 +110,16 @@ def _retrieve_possible_default_config_dirs(self):
...
         possible_default_yml_dirs = self._retrieve_possible_default_config_dirs()
         for default_yml_dir in possible_default_yml_dirs:
-            default_yml_filepath = default_yml_dir / DEFAULTS_FILE_NAME
-            if default_yml_filepath.exists():
-                default_yaml_filepaths.append(default_yml_filepath)
+            for default_file_name in DEFAULTS_FILE_NAMES:
+                default_yml_filepath = default_yml_dir / default_file_name
+                if default_yml_filepath.exists():
+                    default_yaml_filepaths.append(default_yml_filepath)
+                    break  # Only use the first one found (yml preferred over yaml)
```
pankajkoti, line 122: Can we user-facingly document this behaviour that yml is preferred over yaml? Otherwise this remains hidden in the implementation silently

pankajkoti (approved):
LGTM. But would be nice to document the preference of yml over yaml when both are provided

dagfactory/dagfactory.py
pankajastro: Added in commit: 01ae14e7d18f0cd0ccf821ce5352ff8f46302695)

Issue #481, opened 2025-07-16
[Bug] Fix the docs deploy job


Logicnosh:
### DAG Factory version

0.19.0

### airflow version

2.9.0

### Python version

3.10

### Deployment

Other 3rd-party Helm chart

### What happened?

A bug happened!

### Relevant log output

```shell

```

### How to reproduce

the issues is kept facing the error

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

Issue #470, opened 2025-07-09
Fix the docs deploy job


pankajastro:
After removing the pull_request_target event and introducing the Authorization job, it seems the Deploy Pages CI job is now broken.
PR: https://github.com/astronomer/dag-factory/pull/465
CI: https://github.com/astronomer/dag-factory/actions/runs/16163873886/job/45621139666?pr=465

PR #568, opened 2025-08-25
CI: Only build docs on PR push event and deploy for merge and release

pankajastro:
After removing the pull_request_target event and introducing the Authorize job, the Deploy-Pages CI job no longer runs as expected.

This PR fixes that by:
- Ensuring the Deploy-Pages job runs only on merges to main or when a release is published.
- Running the Build-MkDocs job on every push to perform a sanity check on the documentation build.

CI: https://github.com/astronomer/dag-factory/actions/runs/17207460971/job/48811112735

closes: https://github.com/astronomer/dag-factory/issues/481 
closes: https://github.com/astronomer/dag-factory/issues/470

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -256,8 +256,10 @@ jobs:
     needs:
       - Authorize
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v5
+      - uses: actions/checkout@v5
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/cicd.yaml
```diff
@@ -3,7 +3,7 @@ name: CI jobs
 on:
   # run on pushes to the default branch
   push:
-    branches: [main]
+    branches: [main, fix_docs_job]
```
```suggestion
    branches: [main]
```

PR #569, opened 2025-08-25
Update README

pankajastro:
This PR updates the feature section and highlights the Airflow 3 support and DAG-Factory 1.0 release

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,14 +19,19 @@ The minimum requirements for **dag-factory** are:
 For a gentle introduction, please take a look at our [Quickstart Guide](https://astronomer.github.io/dag-factory/latest/getting-started/quick-start-airflow-standalone/). For more examples, please see the
 [examples](/examples) folder.
 
-- [Quickstart](https://astronomer.github.io/dag-factory/latest/getting-started/quick-start-astro-cli/)
+- Quickstart
+    - [Astro CLI](https://astronomer.github.io/dag-factory/latest/getting-started/quick-start-astro-cli/)
+    - [Airflow Standalone](https://astronomer.github.io/dag-factory/latest/getting-started/quick-start-airflow-standalone/)
 - [Benefits](#benefits)
-- [Features](https://astronomer.github.io/dag-factory/latest/features/dynamic_tasks/)
+- Features
+    - [load_yml_dags Function](https://astronomer.github.io/dag-factory/latest/configuration/load_yaml_dags/)
+    - [Schedule](https://astronomer.github.io/dag-factory/latest/configuration/schedule/)
     - [Dynamically Mapped Tasks](https://astronomer.github.io/dag-factory/latest/features/dynamic_tasks/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,14 +19,19 @@ The minimum requirements for **dag-factory** are:
...
     - [Dynamically Mapped Tasks](https://astronomer.github.io/dag-factory/latest/features/dynamic_tasks/)
-    - [Multiple Configuration Files](https://astronomer.github.io/dag-factory/latest/features/multiple_configuration_files/)
-    - [Callbacks](https://astronomer.github.io/dag-factory/latest/features/callbacks/)
+    - [Define Python Object](https://astronomer.github.io/dag-factory/latest/configuration/custom_py_object/)
     - [Custom Operators](https://astronomer.github.io/dag-factory/latest/features/custom_operators/)
-    - [HttpSensor](https://astronomer.github.io/dag-factory/latest/features/http_task/)
+    - [Defaults](https://astronomer.github.io/dag-factory/latest/configuration/defaults/)
+    - [Callbacks](https://astronomer.github.io/dag-factory/latest/features/callbacks/)
+    - [KubernetesPodOperator](https://astronomer.github.io/dag-factory/latest/features/kpo/)
 - [Contributing](https://astronomer.github.io/dag-factory/latest/contributing/howto/)
```
pankajkoti, line 35: I think we're repeating some of the sections here and in the user facing documentations. Can we just add link to user facing documentation in the begining and remove the ones from here that are already in the user facing public documentation and keep only those that are not in there. This makes maintenance easier than having to add/update in both places

README.md
```diff
@@ -36,6 +41,16 @@ For a gentle introduction, please take a look at our [Quickstart Guide](https://
...
+## 📢 Dag-Factory 1.0 Released
+
+Version **1.0** introduces important improvements and breaking changes to support modern Airflow usage.
+
+👉 See the [Migration Guide](https://astronomer.github.io/dag-factory/latest/migration_guide/) to upgrade from earlier versions.
+
+## 🚀 Dag-Factory Supports Airflow 3
+
+DAG-Factory is compatible with **Apache Airflow 3** and supports modern scheduling, and updated import paths.
+
```
pankajkoti, line 53: Do we need this section repeated here? We have it already in the user facing public documentation.

Would be nice to not repeat information to avoid maintenance efforts.

README.md
pankajastro: I feel this might require some effort to automate, so that we maintain a single README and reuse it in docs/index.md as well.
Alternatively, we could remove the "Features" section from the README entirely and just include a link to the full documentation, and I'll log an issue to fully automate in future

What would you suggest?

README.md
pankajastro: I added these two points temporarily to highlight them, and we can remove them in the next release, maybe in 1.1

tatiana (approved):
I agree with @pankajkoti that there is some redundancy, but I'm also happy for us to ship this as it is. I believe both the 1.0 announcement and the AF3 support should not live in the README, but I can also understand the idea of having them there temporarily.

PR #571, opened 2025-08-25
Document dag-factory reserved keys

pankajastro:
Document Reserved Keys

DAG Factory reserves certain YAML keys for internal use. These reserved keys are:

- `__type__`
- `__args__`
- `__join__`
- `__and__`
- `__or__`

### DIFF
--- a/docs/configuration/configuring_workflows.md
+++ b/docs/configuration/configuring_workflows.md
@@ -39,6 +39,16 @@ The **dictionary format** is also supported for backward compatibility. In this 
 !!! note "Format Recommendation"
     While both formats are supported, **we recommend using the list format** as it is more readable and easier to maintain.
 
+## Reserved Keys
+
+DAG Factory reserves certain YAML keys for internal use. These reserved keys are:
+
+- `__type__`
+- `__args__`
+- `__join__`
+- `__and__`
+- `__or__`
+
 ### Check out more configuration params
 
 - [Environment variables](environment_variables.md)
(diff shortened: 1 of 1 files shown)


docs/configuration/configuring_workflows.md
```diff
@@ -39,6 +39,16 @@ The **dictionary format** is also supported for backward compatibility. In this
 !!! note "Format Recommendation"
     While both formats are supported, **we recommend using the list format** as it is more readable and easier to maintain.
 
+## Reserved Keys
+
+DAG Factory reserves certain YAML keys for internal use. These reserved keys are:
```
pankajkoti: Can we elaborate a bit on what reserved means here and what are the implications? I mean why users need to care about these reserved keys or what care do they need to take about these reserved keys? And can we point to some examples here if users need to use these reserved keys in their YAMLs?

pankajastro: The user can use these as keys in their YAML DAG, but only for a specific purpose. 
For example, 
- If we see key `__and__` in YAML, we try to resolve that as a logical expression.
- If we key `__type__` we try to resolve parent as Python object

So if the user has YAML like below, they may encounter an issue 
```
__and__: d
```

tatiana (approved):

docs/configuration/configuring_workflows.md
pankajastro: Added more text to reflect the behaviour

PR #576, opened 2025-09-02
Fix require response_check param in http sensor

### LINKED ISSUE #574: Ability to skip response_check params with HttpSensor
I have a sensor which extends HttpSensor and takes a few params. It fully handles the response and succeeds or fails appropriately. 

I run it like:

```
tasks:
  do-something-task:
    operator: a.b.DoSomethingSensor
    name: foo
    action: bar
    handler: baz
```

This complains that I didn't provide the right combination of `response_check_*` parameters. It would be great if this was not enforced for classes that extend HttpSensor as I do not want a response check passed in.

pankajastro:
The response_check parameter is optional in the HttpSensor, so dag-factory should not make it a required parameter.
https://github.com/apache/airflow/blob/main/providers/http/src/airflow/providers/http/sensors/http.py#L110

closes: https://github.com/astronomer/dag-factory/issues/574

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -217,17 +217,6 @@ class DagBuilder:
                     headers["Content-Type"] = "application/json"
                 task_params["headers"] = headers
         elif HTTP_SENSOR_CLASS and issubclass(operator_obj, HTTP_SENSOR_CLASS):
-            if not (
-                task_params.get("response_check_name") and task_params.get("response_check_file")
-            ) and not task_params.get("response_check_lambda"):
-                raise DagFactoryException(
-                    "Failed to create task. HttpSensor requires \
-                    `response_check_name` and `response_check_file` parameters \
-                    or `response_check_lambda` parameter."
-                )
-
-            # remove dag-factory specific parameters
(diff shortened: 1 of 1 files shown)


tatiana (approved):

dagfactory/dagbuilder.py
```diff
@@ -217,17 +217,6 @@ def _handle_http_sensor(operator_obj, task_params):
...
-                task_params.get("response_check_name") and task_params.get("response_check_file")
-            ) and not task_params.get("response_check_lambda"):
-                raise DagFactoryException(
-                    "Failed to create task. HttpSensor requires \
-                    `response_check_name` and `response_check_file` parameters \
-                    or `response_check_lambda` parameter."
-                )
-
-            # remove dag-factory specific parameters
-            # Airflow 2.0 doesn't allow these to
```
pankajkoti, line 230: Were we able to test this? I am worried if the else block below might cause a problem for the lambda evaluation. Would be nice to confirm the behavior once end to end with a DAG if not already tested

pankajastro: Aha, right address in PR: https://github.com/astronomer/dag-factory/pull/577

PR #577, opened 2025-09-03
Add none check before evaluating lambda func

pankajastro:
Address: https://github.com/astronomer/dag-factory/pull/576#discussion_r2316397605

This PR add a check before evaluating the lambda function in http sensor


pankajkoti (approved):
Makes sense, thank you for fixing this!

PR #585, opened 2025-09-24
Restrict the hatch version to fix the CI

pankajastro:
This PR updates the CI workflow to fix installation errors caused by `click==8.3.0`.

- Replaces the command `uv tool install hatch` with:
  ```bash
  uv pip install --system "hatch>=1.14.2"

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -3,7 +3,7 @@ name: CI jobs
 on:
   # run on pushes to the default branch
   push:
-    branches: [main, fix_ci]
+    branches: [main]
   # run on pull requests originated from forks based on the `main` branch.
   # note that if you're trying to edit the CI in a pull request,
   # your changes won't run here. you need to temporarily add your branch to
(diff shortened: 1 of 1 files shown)


.github/workflows/cicd.yaml
```diff
@@ -3,7 +3,7 @@ name: CI jobs
 on:
   # run on pushes to the default branch
   push:
-    branches: [main]
+    branches: [main, fix_ci]
```
```suggestion
    branches: [main]
```

.github/workflows/cicd.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
           python-version: "3.12"
           architecture: "x64"
 
-      - run: uv tool install hatch
+      - run: uv pip install --system "hatch" "click!=8.3.0"
```
pankajkoti:
```suggestion
      - run: uv pip install --system "hatch>=1.14.2"
```

Should we instead try this? Based on https://github.com/pypa/hatch/issues/2050

.github/workflows/cicd.yaml
```suggestion
    branches: [main]
```

.github/workflows/cicd.yaml
pankajastro: yes, thanks

pankajkoti (approved):

PR #586, opened 2025-09-24
Remove `sla_miss_callback` from dag_kwargs for Airflow >= 3.1.0

jroachgolf84:
I was notified of the following issue. This PR aims to fix that:

> This might be expected, but I wanted to check if you're aware of it. I tried dag-factory 1.0.0 with the latest Airflow 3.1 nightly :airflow_spin:.
> 
> Even a simple DAG definition fails with:
> 
> ```
> TypeError: DAG.__init__() got an unexpected keyword argument 'sla_miss_callback'
> The issue is that dagbuilder.py sets the parameter:
> dag_kwargs["sla_miss_callback"] = dag_params.get("sla_miss_callback", None)
> ```
> 
> every time a DAG is instantiated. The parameter was [deprecated in 3.0](https://airflow.apache.org/docs/task-sdk/stable/api.html#airflow.sdk.DAG) and removed in 3.1. I assume this will be addressed once 3.1 is released, but I couldn't find an open issue yet, so I wanted to flag it.

### DIFF
--- a/tests/test_dagbuilder.py
+++ b/tests/test_dagbuilder.py
@@ -802,7 +802,7 @@ def test_make_dag_with_callbacks():
         from airflow.providers.slack.notifications.slack import send_slack_notification
 
         dag_config_callbacks__with_provider = dict(DAG_CONFIG_CALLBACKS)
-        dag_config_callbacks__with_provider["sla_miss_callback"] = {  # TODO: Deprecate this
+        dag_config_callbacks__with_provider["sla_miss_callback"] = {
             "callback": "airflow.providers.slack.notifications.slack.send_slack_notification",
             "slack_conn_id": "slack_conn_id",
             "text": f"""Sample callback text.""",


tests/test_dagbuilder.py
```diff
@@ -7,6 +7,7 @@
 
 import pendulum
 import pytest
+from airflow.www.views import Airflow
```
kaxil: Why is this needed?

jroachgolf84: Hmmm, not sure how that snuck in there. Maybe pre-commit? I'll remove that now.

dagfactory/dagbuilder.py
```diff
@@ -140,6 +140,10 @@ def get_dag_params(self) -> Dict[str, Any]:
             "on_skipped_callback",  # Not applicable at the DAG-level
             "sla_miss_callback",  # Not applicable at the default_args level
         ]:
+            if callback_type == "sla_miss_callback" and version.parse(AIRFLOW_VERSION) >= version.parse("3.1.0"):
+                # sla_miss_callbacks are fully-deprecated as of 3.1.0. No need to raise a warning, we can silently skip
```
tatiana, line 144: We should log a  warning here, instead of silently passing

tatiana (approved):
Thanks for fixing this, @jroachgolf84 ! I'll wait to merge when we are in a IST friendly time so @pankajastro can review it before we merge.

tests/test_dagbuilder.py
```diff
@@ -797,7 +797,8 @@ def test_make_dag_with_callbacks():
     assert callable(dag.on_failure_callback)
     assert dag.on_failure_callback.__name__ == "print_context_callback"
 
-    if version.parse(AIRFLOW_VERSION) >= version.parse("2.6.0"):
+    # sla_miss_callback is fully deprecated as of Airflow 3.1.0
```
pankajastro, line 800: Deprecated or removed?

pankajastro (approved):
LGTM

dagfactory/dagbuilder.py
pankajastro: Yeah, logging may be a good idea here

pankajkoti (commented):
LGTM. But @pankajastro, I think we need a fix on how we evaluate and set `dag_kwargs`. If the user is not setting `sla_miss_callback` and other such params in the YAML, I believe we should not add it to the kwargs at all. If the parameters exist in the Airflow DAG init, let Airflow set the default values instead of us setting the defaults to `None`.

dagfactory/dagbuilder.py
pankajastro: We will address this in a separate PR

tests/test_dagbuilder.py
pankajastro: We will address this in a separate PR

PR #593, opened 2025-09-30
feat: support `key: +{task}[{xcom_key}]` use

yanshil:
Airflow supports use of {decorated_task}[{xcom_key}] for xcoms that are not return_value [link](https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html#the-big-picture-a-taskflow-pipeline)

Here I support for fetching upstream task's xcom from decorated task with `multiple_outputs=True`

The following example has already been tested in Airflow 3.1.0 (and should also works in Airflow 3)

```
example_multiple_outputs_taskflow:
  schedule: null
  render_template_as_native_obj: True
  catchup: false
  tags: [example, test]
  tasks:
    - task_id: collect
      multiple_outputs: true
      decorator: airflow.decorators.task
      python_callable: sample.collect
    - task_id: echo1
      decorator: airflow.decorators.task
      python_callable: sample.echo
      value: "+collect['key1']"
    - task_id: echo2
      decorator: airflow.decorators.task
      python_callable: sample.echo
      value: "+collect['key2']"
```

### DIFF
--- a/dev/dags/sample.py
+++ b/dev/dags/sample.py
@@ -10,6 +10,19 @@ except ImportError:
     from airflow.operators.python import get_current_context
 
 
+from airflow.utils.context import Context
+
+
+def collect(**context: Context):
+    return {"key1": "value1", "key2": "value2"}
+
+
+def echo(value):
+    print(value)
+    return value
+
+
+
 def build_numbers_list():
     return [2, 4, 6]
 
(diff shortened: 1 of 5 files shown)


dev/dags/sample.py
```diff
@@ -10,6 +10,19 @@
     from airflow.operators.python import get_current_context
 
 
+from airflow.utils.context import Context
+
+
+def collect(**context: Context):
```
pankajastro, line 16: Where are you using this?

dev/dags/sample.py
```diff
@@ -10,6 +10,19 @@
...
+    return {"key1": "value1", "key2": "value2"}
+
+
+def echo(value):
```
pankajastro, line 20: Where are you using this?

pankajastro (commented):
Could you please add an example DAG [here](https://github.com/astronomer/dag-factory/tree/main/dev/dags) and document this feature?

pankajastro (approved):

PR #599, opened 2025-10-15
Support dataset timetable

AppKiv:
support DatasetOrTimeSchedule
add processing key "timetable" for schedule with key "datasets"

example:
```
  schedule:
    datasets: "((s3://dataset_custom_1 & s3://dataset_custom_2) | s3://dataset_custom_3)"
    timetable: 
      __type__: airflow.timetables.trigger.CronTriggerTimetable
      cron: "0 1 * * 5"
      timezone: UTC
```


dev/dags/datasets/example_dataset_or_timetable_schedule.yml
```diff
@@ -0,0 +1,16 @@
+consumer_time_schedule_dag:
+  catchup: false
+  default_args:
+    owner: "example_owner"
+    start_date: '2024-01-01'
+  description: "Example DAG consumer scheduled by datasets or timetable"
+  schedule:
+    datasets: "((s3://bucket-cjmm/raw/dataset_custom_1 & s3://bucket-cjmm/raw/dataset_custom_2) | s3://bucket-cjmm/raw/dataset_custom_3)"
```
pankajastro: Can you use a local file instead of S3? A local file would make it easier to validate in CI and avoid any future regression.

AppKiv: Changed to abstract URI,  originally just copied it from other examples.
URI is an abstraction in Airflow; it doesn’t have to be an S3 path or a local file path.

dagfactory/dagbuilder.py
```diff
@@ -616,8 +622,19 @@ def configure_schedule(dag_params: Dict[str, Any], dag_kwargs: Dict[str, Any]) -
...
+                    if has_timetable_attr:
+                        timetable_schedule = schedule["timetable"]
+                        if version.parse(AIRFLOW_VERSION) < version.parse("3.0.0"):
+                            dag_kwargs[schedule_key] = DatasetOrTimeSchedule(
+                                timetable=timetable_schedule, datasets=datasets_schedule
+                            )
+                        else:
+                            dag_kwargs[schedule_key] = AssetOrTimeSchedule(
+                                timetable=timetable_schedule, assets=datasets_schedule
+                            )
```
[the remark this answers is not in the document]
AppKiv: added test for AF3

pankajastro (approved):

PR #600, opened 2025-10-15
Drop Python 3.9 Support

pankajastro:
Drop the Python 3.9 Support 

- Python 3.9 will reach EOL at the end of this month: https://www.python.org/downloads/
- Airflow has already dropped Python 3.9 in the latest release


kaxil (approved):
We should add support for Py 3.13

PR #613, opened 2025-11-14
fix: use __file__-relative paths in dev examples for Airflow 3 on Astro Hosted

pankajastro:
When running Airflow 3 on Astro Hosted with DAG Versioning (https://www.astronomer.io/docs/astro/dag-versioning) enabled, DAGs are packaged into versioned bundles. In this model, hardcoded absolute paths (e.g. /usr/local/airflow/dags/) and environment variable fallbacks (CONFIG_ROOT_DIR) are not reliable — the bundle's location on disk is managed by Astro and not guaranteed to match the old path assumption.                                        
                                                                                                                                                         
  Using Path(__file__).resolve().parent resolves the YAML config path relative to the DAG file itself, which works correctly regardless of where Astro     
  places the bundle.
                                                                                                                                                           
**Changes**                                                                                                                           

  - Replace os.getenv("CONFIG_ROOT_DIR", "/usr/local/airflow/dags/") with Path(__file__).resolve().parent across all dev/dags/ example loaders             
  - Bump dev Dockerfile base image to astrocrpublic.azurecr.io/runtime:3.1-5
                                                                                                                                                           
**Test plan**                                                                                                                    
                                                                                                                                                           
  - Deploy to Astro Hosted with DAG Versioning enabled and verify all example DAGs parse and load correctly

### DIFF
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -1,5 +1,5 @@
 # FROM quay.io/astronomer/astro-runtime:12.7.0
-FROM astrocrpublic.azurecr.io/runtime:3.0-4
+FROM astrocrpublic.azurecr.io/runtime:3.1-5
 
 ENV CONFIG_ROOT_DIR=/usr/local/airflow/dags/
 ENV PYTHONPATH=$AIRFLOW_HOME/dags:$PYTHONPATH
@@ -11,6 +11,7 @@ RUN apt-get update && apt-get install -y jq
 # Install uv for faster package management
 RUN pip install uv
 
+# RUN uv pip install dag-factory #--system /usr/local/airflow/include/*.whl
 RUN uv pip install --system /usr/local/airflow/include/*.whl
 
 USER astro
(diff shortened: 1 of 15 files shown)


tatiana (approved):
Looks great, Pankaj!
I know this is in draft, but it solves an importante problem.
Please, update the title/description so it says it is a fix related to running Airflow 3 in Astro Hosted.
The description should explain this is due to how Astro DAG Versioning Bundles work. More details can be found in:
https://www.astronomer.io/docs/astro/dag-versioning

dev/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN apt-get update && apt-get install -y jq
 # Install uv for faster package management
 RUN pip install uv
 
+# RUN uv pip install dag-factory #--system /usr/local/airflow/include/*.whl
```
pankajkoti:
```suggestion
```
Do we need the comment?

dev/dags/example_dag_factory.yml
```diff
@@ -5,7 +5,7 @@ default:
 # ----8<---   [ start: example_dag_yaml_configuration ]
 basic_example_dag:
   default_args:
-    owner: "custom_owner"
+    owner: "custom_owner1"
```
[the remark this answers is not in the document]
pankajastro: Fixed

dev/Dockerfile
pankajastro: Fixed

PR #656, opened 2026-02-10
Remove structlog pin

### LINKED ISSUE #609: Remove `structlog` dependency for docs CI job
PR: https://github.com/astronomer/dag-factory/pull/608 add structlog as a dependency to prevent docs job failures

We should remove this `structlog` as a dependency once Apache Airflow 3.1.2 is released and available, as that version is expected to resolve the compatibility issue.

pankajkoti:
closes: #609

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -165,9 +165,6 @@ dependencies = [
     "mike",
     "pymdown-extensions",
     "mkdocs-material",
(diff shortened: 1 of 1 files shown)


tatiana (approved):
Thanks, @pankajkoti , always good to do some house clean up!

PR #697, opened 2026-04-06
Ignore more editor local settings folder

gyli:
Add Cursor and Claude settings folders to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,6 +106,12 @@ pip-selfcheck.json
 # vscode
 .vscode
 
+# Claude Code
+.claude/
+
+# Cursor
+.cursor/
+
 # IntelliJ
 .idea/*
 
--- a/.gitignore
+++ b/.gitignore
@@ -107,10 +107,12 @@ pip-selfcheck.json
 .vscode
 
 # Claude Code
-.claude/
+.claude/settings.local.json
+.claude/plans/
+CLAUDE.local.md
 
 # Cursor
-.cursor/
+.cursor/plans/
 
 # IntelliJ
 .idea/*
--- a/dev/simple_auth_manager_passwords.json.generated
+++ b/dev/simple_auth_manager_passwords.json.generated
@@ -1,0 +1,1 @@
+{"admin": "mapvKkZ5Vnexq8sN"}
--- a/dev/simple_auth_manager_passwords.json.generated
+++ b/dev/simple_auth_manager_passwords.json.generated
@@ -1,1 +1,0 @@
-{"admin": "mapvKkZ5Vnexq8sN"}


pankajkoti (commented):
We may not actually want to ignore these folders. In fact, if you have created files in this folder that you can help other devs that include cursor rules or claude markdown files, it would be encouraged to contribute such files to the repo @gyli :).

gyli:
Hi @pankajkoti We might need a broader discussion about how we want to persist the agent rules, skills, guideline, etc. for this repo. 

I do have a plan to create agent skills for dag-factory https://github.com/astronomer/dag-factory/issues/687, but I am not sure yet if `.cursor` or `.claude` would be the preferred location for it. Here, I am trying to follow how it's handled in the Airflow repo, which ignores these folders, and keeps AGENTS.md at the root level. While Astro also has a separate repo for Airflow AI plugin/skills https://github.com/astronomer/agents/tree/main/skills.

On the other hand, some local setting files under these folders should be ignored for sure, like `.claude/settings.local.json`. 

My suggestion is to create a tool-agnostic file like `AGENTS.md` and folder like `.agents` or `.ai` for skills, and symlink it to `.cursor` or `.claude`. If you think it's good, I will update this PR.

pankajkoti:
> Hi @pankajkoti We might need a broader discussion about how we want to persist the agent rules, skills, guideline, etc. for this repo.
> 
> I do have a plan to create agent skills for dag-factory #687, but I am not sure yet if `.cursor` or `.claude` would be the preferred location for it. Here, I am trying to follow how it's handled in the Airflow repo, which ignores these folders, and keeps AGENTS.md at the root level. While Astro also has a separate repo for Airflow AI plugin/skills https://github.com/astronomer/agents/tree/main/skills.
> 
> On the other hand, some local setting files under these folders should be ignored for sure, like `.claude/settings.local.json`.
> 
> My suggestion is to create a tool-agnostic file like `AGENTS.md` and folder like `.agents` or `.ai` for skills, and symlink it to `.cursor` or `.claude`. If you think it's good, I will update this PR.

Thanks for the thoughtful proposal @gyli! Here are my thoughts:

I like the idea of having an `AGENTS.md` at the root level as a tool-agnostic starting point.
However, I'd keep it simple and avoid the symlink approach. Symlinks can be fragile, and each tool has its own native format, so it might still help to have those.

My suggestion:
 - Don't blanket-ignore .cursor/.claude folders — shareable configs (project-level settings, rules) benefit all contributors
 - Do selectively ignore personal/local files (e.g., .claude/settings.local.json)
 - Do add an AGENTS.md at the root as tool-agnostic guidance

gyli:
@pankajkoti PR updated

pankajkoti (approved):

PR #703, opened 2026-04-10
Simplify DAG kwargs construction with a declarative spec

### LINKED ISSUE #587: Remove default value for `dag_kwargs` params
For many dag_kwargs parameters, we currently set default values, mostly None. In DAG Factory, we should avoid setting these defaults ourselves and instead rely on Airflow to handle them.

https://github.com/astronomer/dag-factory/pull/586#pullrequestreview-3266103199

gyli:
Resolves https://github.com/astronomer/dag-factory/issues/587

Key changes:
1. This PR uses dict-based declarative spec defining DAG arguments, achieving unified argument definition like version check.
2. Always fall back to Airflow's own default value or behavior when key not set.

Future works: task-level spec will be added in the follow-up PR

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -781,7 +781,6 @@ class DagBuilder:
         explicitly sets a parameter that is incompatible with the installed version.
         """
         dag_kwargs: Dict[str, Any] = {}
-        installed = INSTALLED_AIRFLOW_VERSION
 
         for spec in _DAG_PARAM_SPEC:
             key: str = spec["key"]
@@ -796,18 +795,18 @@ class DagBuilder:
             min_ver = spec.get("min_version")
             max_ver = spec.get("max_version")
 
-            if min_ver and installed < min_ver:
+            if min_ver and INSTALLED_AIRFLOW_VERSION < min_ver:
                 warnings.warn(
                     f"DAG parameter '{key}' requires Airflow >= {min_ver} "
-                    f"(installed: {AIRFLOW_VERSION}). The parameter will be ignored.",
+                    f"(installed: {INSTALLED_AIRFLOW_VERSION}). The parameter will be ignored.",
                     UserWarning,
                 )
                 continue
 
-            if max_ver and installed >= max_ver:
+            if max_ver and INSTALLED_AIRFLOW_VERSION >= max_ver:
                 warnings.warn(
                     f"DAG parameter '{key}' is not supported in Airflow >= {max_ver} "
-                    f"(installed: {AIRFLOW_VERSION}). The parameter will be ignored.",
+                    f"(installed: {INSTALLED_AIRFLOW_VERSION}). The parameter will be ignored.",
                     UserWarning,
                 )
                 continue
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -13,7 +13,6 @@ from datetime import datetime
 from functools import partial
 from typing import Any, Callable, Dict, List, Tuple, Union
 
-from airflow import configuration
 from packaging import version
 
 from dagfactory.utils import check_dict_key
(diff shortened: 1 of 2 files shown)


dagfactory/dagbuilder.py
```diff
@@ -79,6 +78,46 @@
...
+# Declarative spec for DAG-level parameters passed to the DAG constructor.
+# Keys:
+#   key                    - the DAG constructor kwarg name (also the dag_params lookup key)
+#   required               - raise DagFactoryConfigException if absent
+#   min_version            - only applicable when Airflow >= this version (Version object); warn+skip if param is set below this
+#   max_version            - only applicable when Airflow <  this version (Version object); warn+skip if param is set at/above this
+#   deprecated_in_favor_of - this key is deprecated; when present, warn and write the value to the given canonical key.
+#                            must appear before the canonical key in the spec so the canonical key wins when both are set.
+#                            warning message is auto-generated from key and deprecated_in_favor_of.
+#   transform              - optional module-level callable applied to the raw dag_params value before writing to dag_kwargs.
```
pankajkoti, line 90: I don't see an example for this currently. Do we need this now? or can we add it later when we have the use case?

dagfactory/dagbuilder.py
```diff
@@ -730,88 +770,80 @@ def _normalise_task_groups_config(task_groups_cfg: Any) -> Dict[str, Dict[str, A
...
+                )
+                continue
 
-        if version.parse(AIRFLOW_VERSION) < version.parse("3.1.0"):
-            # sla_miss_callback is fully-deprecated as of 3.1.0
-            dag_kwargs["sla_miss_callback"] = dag_params.get("sla_miss_callback", None)
+            deprecated_in_favor_of: str = spec.get("deprecated_in_favor_of", "")
+            transform = spec.get("transform")
+            value = dag_params[key]
+            value = transform(value) if transform else value
```
pankajkoti, line 817: Should we defer this to a later PR when we have the use case? It looks like this code won't be used.

pankajkoti (approved):
Looks solid to me @gyli . Apologies for the delay in reviewing here. Thanks for proposing this. Looks like we have some conflicts with the main branch. Can you please rebase/merge to resolve the conflicts?


@pankajastro  / @tatiana, would you like to give another pair of eyes for this simplification interface?

gyli:
This PR has a potential conflict with my another PR https://github.com/astronomer/dag-factory/pull/732 
Both of these PRs specify param attributes like `min_version`, `max_version` and more, one in jsonschema, and one in dict-based spec. They are similar but serve for different purposes - jsonschema is nice to have for linting as it provides IDE-native validation; Python dict-based spec is more suitable for DAG building which supports Python func/object for value transformation.

@pankajkoti  @pankajastro I would also love to hear your suggestions on this.

dagfactory/dagbuilder.py
gyli: `transform` can be used for replacing `configure_schedule`, as mentioned in the TODO at line 615.

dagfactory/dagbuilder.py
gyli: Plz see the reply in the other comment

pankajkoti:
> This PR has a potential conflict with my another PR #732 Both of these PRs specify param attributes like `min_version`, `max_version` and more, one in jsonschema, and one in dict-based spec. They are similar but serve for different purposes - jsonschema is nice to have for linting as it provides IDE-native validation; Python dict-based spec is more suitable for DAG building which supports Python func/object for value transformation.
> 
> @pankajkoti @pankajastro I would also love to hear your suggestions on this.

Thanks @gyli for catching this and bringing the potential conflict forward before either PR landed.

Agree these overlap, and I'd rather not maintain the version/deprecation facts in two places since they'll drift the moment Airflow changes a param (we already have `dag_display_name`, `default_view`, `orientation`, `sla_miss_callback` gates duplicated across both).

I see two ends of the spectrum: keep them fully separate (simple now, drifts later), or fully generate the jsonschema from the Python spec (no drift, but more machinery). Does a middle ground make sense here? Keep one canonical table of param metadata (key + min/max version + deprecation) that both read: the build-time dict spec (#703) layers `transform`/`required` on top, the jsonschema (#732) layers type definitions on top, plus a sync test in CI so the two can't silently diverge without fully generating one from the other.

Since you own both, the easiest path is probably to land one first, then rebase the other onto the shared table. Does that middle option sound right to you?

By shared table, I mean something like
```
# param_metadata.py  -- single source of truth for per-param version/deprecation facts
DAG_PARAM_METADATA = {
    "dag_display_name": {"min_version": "2.9.0"},
    "default_view":     {"max_version": "3.0.0"},
    "orientation":      {"max_version": "3.0.0"},
    "sla_miss_callback":{"max_version": "3.1.0"},
    "concurrency":      {"deprecated_in_favor_of": "max_active_tasks"},
    ...
}
```
Then each PR reads from it instead of restating the numbers:

- #703 builds its _DAG_PARAM_SPEC from this table and adds the build-only bits on top (transform, required) that can't live in JSON.
- #732 generates (or, at minimum, validates) its x-airflow-* annotations against this table.

gyli:
@pankajkoti

Sorry for the late reply - 
I also hate to maintain similar params metadata in 2 places, and I am evaluating these 2 potential approaches:
1. jsonschema-based schema for both linting and dag building. Linting is a required step in dag building. This approach allows in-IDE YAML validation through the jsonschema file. While Python object still needs to be appended to some params for value transformation.
2. Python dict based schema for linting and dag building.

Overall, these 2 approaches have similar complexities and parsing performance. The main difference seems to be supporting in-IDE YAML validation or not. Still working on the comparison and will share my result later.

PR #732, opened 2026-04-26
Add jsonschema to lint command

### LINKED ISSUE #87: Add json schema validation for yaml files

gyli:
Resolves: https://github.com/astronomer/dag-factory/issues/87

This PR enhances the `dagfactory lint` command, to not only validate if YAML syntax is correct, but also validate if it's a valid DAG config.

Key changes:
1. It accepts linting py, yaml, and dir. The py file is recommended for the full support of all kinds of defaults.
2. It adds jsonschema defining dag-level parameters, which can be used for performing fast validating YAML without calling Airflow. It can also be used in IDE for YAML validation without calling Python.


dagfactory/validator.py
```diff
@@ -0,0 +1,520 @@
...
+            result.issues.append(ValidationIssue(
+                file=source, dag_id=None, severity="warning",
+                message="No DAG entries found (top-level values are not mappings); "
+                "looks like a config-only YAML, skipping.",
+            ))
+            return result
+
+        if not self.schema_only:
+            try:
+                _DagFactory(config_dict=config).build_dags()
```
tatiana: I believe this breaks once we rebase on `main`. There, `build_dags()` catches per-DAG errors and returns `(dags, first_build_error)` rather than raising (it only re-raises when `strict_mode` is on, which is off by default), so this `except` won't fire and we'd report invalid DAGs as OK. Could we read `first_build_error` from the return value (or turn on `strict_mode` for the lint run) instead of relying on the exception? The loader path at line 388 has the same assumption. WDYT?

tests/test__main__.py
```diff
@@ -57,46 +88,175 @@ def test_help_option():
...
+    assert "config.yml" in rendered
+
+
+def test_lint_single_non_loader_py_is_skipped(tmp_path):
+    """Single explicit .py target that doesn't import dagfactory — visible skip."""
+    target = tmp_path / "utils.py"
+    target.write_text("def helper():\n    return 42\n")
+    result = runner.invoke(app, ["lint", str(target)])
+    assert result.exit_code == 0
+    assert "does not import dagfactory" in result.stdout
```
tatiana, line 242: Heads-up: this fails when the console width is 80, which is what Rich falls back to with no TTY (i.e. in CI) - the long tmp path wraps and splits `import` / `dagfactory`. It passes on a wide terminal locally. Could we pin the console width in the test, or assert in a way that ignores wrapping?

dagfactory/validator.py
```diff
@@ -0,0 +1,520 @@
...
+
+    original = df_module.DagBuilder
+    df_module.DagBuilder = _Capturing
+    try:
+        yield captured
+    finally:
+        df_module.DagBuilder = original
+
+
+def _import_loader(py_path: Path) -> Tuple[Optional[Any], Optional[Exception]]:
```
tatiana, line 328: I think we should register the module in `sys.modules` before `exec_module` (and pop it after). Otherwise a valid loader that uses `from __future__ import annotations` + a dataclass, `pickle`, or relative imports fails to import and we'd flag it as broken. Using a unique module name per file would help too - we currently reuse `_dagfactory_lint_module` for every file.

dagfactory/validator.py
```diff
@@ -0,0 +1,520 @@
...
+            )]
+        return [self._validate_yaml_config(config, yaml_file_path)]
+
+    def validate_yaml_content(
+        self, yaml_content: str, source_label: str = "<inline yaml>"
+    ) -> List[FileValidationResult]:
+        """Lint an inline YAML string. ``__and__``/``__or__``/``__join__`` directives are not expanded."""
+        source = Path(source_label)
+        try:
+            raw = yaml.load(os.path.expandvars(yaml_content), Loader=yaml.FullLoader)
```
tatiana: This path does `expandvars -> FullLoader -> cast_with_type` but skips the `__join__`/`__and__`/`__or__` flattening that `load_yaml_file` does, so `--yaml-content` validates a different shape than the file path (and than what dag-factory actually builds). Could we pull out a small `load_yaml_string()` helper and share it between both so they stay in sync?

dagfactory/validator.py
```diff
@@ -0,0 +1,520 @@
...
+                            message=f"dag-factory failed to build DAGs: {type(exc).__name__}: {exc}",
+                        )
+                    )
+            for builder in builders:
+                merged = merge_configs(builder.dag_config, builder.default_config)
+                self._validate_dag(target, builder.dag_name, merged, result)
+            results.append(result)
+        return results
+
+    def validate_yaml_file(self, yaml_file_path: Path) -> List[FileValidationResult]:
```
tatiana, line 467: Since we intentionally don't resolve the external `defaults.yml` chain here, a DAG that gets its `start_date` (or `tasks`) from defaults - a common pattern for us - fails as a hard error even though it builds fine. Could we treat `start_date`/`tasks` as warnings in the schema-only path rather than errors? (Same applies to `required: [tasks]` / `x-required-anywhere` in the schema.)

dagfactory/__main__.py
```diff
@@ -80,35 +118,125 @@ def main(
...
+
+    table = Table(
+        title="[bold][medium_purple3]DAG Factory[/medium_purple3][/bold]: Lint Results",
+        show_lines=True,
+    )
     table.add_column("File", style="cyan", no_wrap=True)
     table.add_column("Status", style="bold")
     table.add_column("Error Message", style="red", no_wrap=False, overflow="fold")
 
+    validator = DagParameterValidator(airflow_version=airflow_version, schema_only=schema_only)
```
tatiana: An invalid `--airflow-version` makes `DagParameterValidator` raise `ValueError`, which surfaces to the user as a full stack trace. Similar to what we discussed for the CLI in #539, I believe the experience is much nicer if we catch it and print a friendly message + `raise typer.Exit(2)` - the message itself is already good.

dagfactory/__main__.py
```diff
@@ -29,20 +32,55 @@
...
+        py_loaders = [p for p in py_candidates if imports_dagfactory(p)]
+        if lint_yaml_in_dir:
+            yaml_files = list(path.rglob("*.yml")) + list(path.rglob("*.yaml"))
+            files = py_loaders + yaml_files
+        else:
+            files = py_loaders
+        if not files:
+            extra = "" if lint_yaml_in_dir else " (pass --lint-yaml-in-dir to also include .yml/.yaml files)"
+            console.print(
+                f"[yellow]No lintable files found in '{path}' "
```
tatiana, line 71: For a folder that only has YAML configs and no `.py` loaders, we print "No lintable files found" and exit 0 - so a CI step that used to validate those YAMLs would now pass without checking anything. Should we exit non-zero here, or at least make it a louder warning? WDYT?

tests/test_validator.py
```diff
@@ -0,0 +1,255 @@
...
+    p.write_text("datasets:\n  - name: d1\n    uri: s3://x\n")
+    results = DagParameterValidator(airflow_version="3").validate_yaml_file(p)
+    assert not results[0].errors
+    assert any("config-only" in i.message for i in results[0].warnings)
+
+
+# ---------------------------------------------------------------------------
+# validate_yaml_content
+# ---------------------------------------------------------------------------
+def test_validate_yaml_content_catches_schema_error():
```
tatiana, line 324: The custom `x-*` keywords are the heart of this PR, but only `schedule_interval` exercises them. Could we add a few small tests - that `x-deprecated-since` / `x-dagfactory-supported` produce **warnings** (not errors), and that `x-mutually-exclusive` / `x-required-anywhere` / `x-airflow-min-version` produce **errors**? It'll make these much safer to change later.

tatiana (commented):
@gyli thanks a lot for this 🎉 - resolving #87 has been on our list for a while, and this goes well beyond it. I really like the modular schema and that you documented the IDE integration too.

Before we go further, could you rebase on `main`? The branch has drifted quite a bit, and there's one behaviour change that matters: `build_dags()` no longer raises on a failed DAG - it now returns `(dags, first_build_error)` and only raises under `strict_mode` (off by default). Our default "build mode" relies on that exception to report errors, so after the rebase it would silently pass invalid DAGs. I left an inline note with a suggestion. The rebase will also collide with the `--ignore` / `.airflowignore` support that landed in `lint` upstream - worth deciding whether we carry it forward.

A few other things I'd like to sort before merging (all inline): the single-YAML path can fail valid DAGs that get `start_date` from `defaults.yml`, the inline `--yaml-content` path validates a slightly different shape than the file path, and there's a test that fails under CI's console width. The rest are small and we can iterate. One to keep an eye on (non-blocking): widening the global `object` type-checker (`validator.py:81`) means a non-dict reaching a node with `required`/`properties` would raise an uncaught `TypeError` - not reachable with today's schema, but a bit fragile.

I pulled the branch and ran it against Airflow 3.0.3 to double-check these. Happy to pair on the `build_dags` part since it's a little subtle. Thanks again!

dagfactory/schemas/dag_parameters.json
```diff
@@ -0,0 +1,655 @@
...
+        "allowed_run_types": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "x-airflow-min-version": "3.2"
+        },
+        "deadline": {
+          "description": "Replacement for the removed SLA feature. Not yet wired through dag-factory (see github issue #696).",
+          "x-airflow-min-version": "3",
```
tatiana: `deadline` was introduced in Airflow 3.1, not 3.0 (it's not in the 3.0.3 `DAG` signature), so today a config using it isn't flagged as too-new on 3.0.x.
```suggestion
          "x-airflow-min-version": "3.1",
```

dagfactory/schemas/dag_parameters.json
```diff
@@ -0,0 +1,655 @@
...
+          "minimum": 1
+        },
+        "max_active_runs": {
+          "type": "integer",
+          "minimum": 1
+        },
+        "concurrency": {
+          "type": "integer",
+          "description": "Deprecated alias for max_active_tasks.",
+          "x-deprecated-since": "2.0",
```
tatiana: `concurrency` became the deprecated alias for `max_active_tasks` in 2.2, not 2.0.
```suggestion
          "x-deprecated-since": "2.2",
```

dagfactory/__main__.py
```diff
@@ -80,35 +118,125 @@ def main(
...
-    if total_errors > 0:
-        console.print(f"Analysed {len(files)} files, found [red]{total_errors}[/red] invalid YAML files.")
+    summary = f"Analysed {analysed} file(s)"
+    if total_errors:
+        console.print(
+            f"{summary}, found [red]{total_errors}[/red] file(s) with errors and "
+            f"[yellow]{total_warnings}[/yellow] warning(s)."
+        )
         if not verbose:
             console.print(f"For more details on the errors, run with --verbose.")
```
tatiana:
```suggestion
            console.print("For more details on the errors, run with --verbose.")
```
(no placeholders, so the `f` isn't needed - ruff F541 would flag it)

tatiana (commented):
_(Apologies - this is a duplicate review entry posted by mistake. Please see my full review just above; this one can be ignored.)_

tatiana:
Thanks a lot for working on this, @gyli , I'd love to see it merged.

Please, could you address Copilot's feedback and rebase? We can aim to have a DAG Factory release in July, depending on how quickly we can wrap up the open PRs.

gyli:
Sorry I missed this one, will be on this soon!

dagfactory/validator.py
gyli: Will force strict_mode during validation

dagfactory/validator.py
gyli: Will use the same load func for parsing file and string

dagfactory/__main__.py
gyli: Will do!

docs/features/cli.md
```diff
@@ -26,74 +26,90 @@ dagfactory [OPTIONS]
...
+| -------------------- | -------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| Python loader (`.py`) | `dagfactory lint dags/loader.py`                         | The loader is imported and every `load_yaml_dags(...)` invocation is captured. dag-factory's own defaults handling (defaults.yml chain, `defaults_config_dict`, etc.) runs end-to-end. |
+| YAML file (`.yml`/`.yaml`) | `dagfactory lint dags/my_dag.yml`                  | Each top-level DAG entry is validated as a self-contained config. The file's own `default:` block is applied; no external defaults are merged.     |
+| Inline YAML          | `dagfactory lint --yaml-content "$(cat my_dag.yml)"`     | Same semantics as a YAML file, supplied as a string. Useful for editor / IDE integration.                                                          |
```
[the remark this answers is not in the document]
gyli: outdated

tests/test__main__.py
gyli: console.size updated

dagfactory/__main__.py
gyli: Added a dedicated log and non-zero code for the case dir has YAML but no loaders

gyli:
PR updated with all comments addressed, waiting for the result from CI

gyli:
@tatiana the tests look good, plz review again.

PR #736, opened 2026-04-30
Fix python_callable resolution on Airflow 2 with providers-standard

pankajkoti:
## Summary

- Fixes #679: on Airflow 2.x with `apache-airflow-providers-standard` installed, `airflow.operators.python.PythonOperator` and `airflow.providers.standard.operators.python.PythonOperator` are **different, unrelated classes**. The single `try/except` import in `dagbuilder.py` bound only one of them, so the `issubclass` check in `make_task` missed YAMLs that referenced the other module path. The result was `python_callable_name`/`python_callable_file` never being resolved into `python_callable`, producing `airflow.exceptions.AirflowException: missing keyword argument 'python_callable'` at parse time.
- Now imports both module paths when available and checks `issubclass` against the union (`PYTHON_CALLABLE_CLASSES`). Both YAML import paths work regardless of which providers are installed.
- Adds a parametrized regression test that exercises both `airflow.operators.python.PythonOperator` and `airflow.providers.standard.operators.python.PythonOperator`, with skip guards for AF3 (no core path) and providers-standard not installed.
- Adds a dedicated CI matrix row — `py3.12 / af2.11 / providers-standard==1.9.1` — wired into the unit-tests workflow so this exact combination is always covered. Achieved by adding a `providers-standard` Hatch matrix variable, a `{matrix:providers-standard:}` substitution in `pre-install-commands`, an optional 3rd arg in `scripts/test/pre-install-airflow.sh`, and an `include:` row + computed env name in `cicd.yaml`.

## Verification

Reproduced the bug locally on Airflow 2.11.0 + `apache-airflow-providers-standard==1.9.1` + `apache-airflow-providers-common-compat==1.14.3` (the reported environment), confirmed both module paths fail before the fix and succeed after. Ran the existing `tests/test_dagbuilder.py` suite on this exact combo: 67 pass, 5 skip, 0 regressions.

| Scenario | Result |
|---|---|
| YAML uses `airflow.operators.python.PythonOperator`, before fix | `AirflowException: missing keyword argument 'python_callable'` |
| YAML uses `airflow.operators.python.PythonOperator`, after fix | works |
| Regression test, providers-standard installed | both params PASS |
| Regression test, providers-standard absent | `[providers-standard]` SKIPPED, `[core]` PASS |

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -63,6 +63,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         airflow-version: ["2.9", "2.10", "2.11", "3.0", "3.1", "3.2"]
+        providers-standard-version: [""]
         exclude:
           - python-version: "3.13"
             airflow-version: "2.9"
@@ -82,11 +83,26 @@ jobs:
             airflow-version: "3.0"
           - python-version: "3.14"
             airflow-version: "3.1"
+        # Regression coverage for issue #679: AF 2.11 + apache-airflow-providers-standard 1.9.1.
+        include:
+          - python-version: "3.12"
+            airflow-version: "2.11"
+            providers-standard-version: "1.9.1"
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
           persist-credentials: false
+
+      - name: Compute hatch env name
+        id: hatch-env
+        run: |
+          suffix=""
+          if [ -n "${{ matrix.providers-standard-version }}" ]; then
+            suffix="-${{ matrix.providers-standard-version }}"
+          fi
+          echo "name=tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }}${suffix}" >> "$GITHUB_OUTPUT"
+          echo "suffix=${suffix}" >> "$GITHUB_OUTPUT"
 
       - uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
@@ -94,7 +110,7 @@ jobs:
             ~/.cache/uv
             ~/.cache/pip
             .local/share/hatch/
-          key: unit-${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.airflow-version }}-${{ hashFiles('pyproject.toml') }}-${{ hashFiles('dagfactory/__init__.py') }}
+          key: unit-${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.airflow-version }}${{ steps.hatch-env.outputs.suffix }}-${{ hashFiles('pyproject.toml') }}-${{ hashFiles('dagfactory/__init__.py') }}
 
       - name: Install uv
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
@@ -109,16 +125,16 @@ jobs:
       - name: Install packages and dependencies
         run: |
           uv pip install  --system "hatch>=1.14.2"
-          hatch -e tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }} run pip freeze
+          hatch -e ${{ steps.hatch-env.outputs.name }} run pip freeze
 
       - name: Test DAG Factory against Airflow ${{ matrix.airflow-version }} and Python ${{ matrix.python-version }}
         run: |
-          hatch run tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }}:test-cov
+          hatch run ${{ steps.hatch-env.outputs.name }}:test-cov
 
       - name: Upload coverage to Github
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: coverage-unit-test-${{ matrix.python-version }}-${{ matrix.airflow-version }}
+          name: coverage-unit-test-${{ matrix.python-version }}-${{ matrix.airflow-version }}${{ steps.hatch-env.outputs.suffix }}
           path: .coverage
           include-hidden-files: true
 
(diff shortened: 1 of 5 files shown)


dagfactory/dagbuilder.py
```diff
@@ -32,12 +32,44 @@
...
+    _python_sensor_classes.append(_CorePythonSensor)
 except ImportError:
-    from airflow.operators.python import BranchPythonOperator, PythonOperator
-    from airflow.sensors.python import PythonSensor
+    pass
+
+PythonOperator = _python_operator_classes[0]
+BranchPythonOperator = _branch_python_operator_classes[0]
+PythonSensor = _python_sensor_classes[0]
+PYTHON_CALLABLE_CLASSES = tuple(_python_operator_classes + _branch_python_operator_classes + _python_sensor_classes)
```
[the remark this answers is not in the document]
pankajkoti: addressed

.github/workflows/cicd.yaml
```diff
@@ -109,16 +125,16 @@
       - name: Install packages and dependencies
         run: |
           uv pip install  --system "hatch>=1.14.2"
-          hatch -e tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }} run pip freeze
+          hatch -e ${{ steps.hatch-env.outputs.name }} run pip freeze
```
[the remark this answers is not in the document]
pankajkoti: addressed

.github/workflows/cicd.yaml
```diff
@@ -109,16 +125,16 @@
...
 
       - name: Test DAG Factory against Airflow ${{ matrix.airflow-version }} and Python ${{ matrix.python-version }}
         run: |
-          hatch run tests.py${{ matrix.python-version }}-${{ matrix.airflow-version }}:test-cov
+          hatch run ${{ steps.hatch-env.outputs.name }}:test-cov
```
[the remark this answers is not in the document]
pankajkoti: addressed

pankajkoti:
Test run for the newly introduced matrix entry with standard provider for no future regression verification: https://github.com/astronomer/dag-factory/actions/runs/25153975049/job/73731052095?pr=736#step:8:45

.github/workflows/cicd.yaml
```diff
@@ -3,7 +3,7 @@ name: CI jobs
 on:
   # run on pushes to the default branch
   push:
-    branches: [main]
+    branches: [main, fix-issue-679-python-callable-airflow2]
```
pankajkoti:
```suggestion
    branches: [main]
```

.github/workflows/cicd.yaml
```diff
@@ -63,6 +63,7 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         airflow-version: ["2.9", "2.10", "2.11", "3.0", "3.1", "3.2"]
+        providers-standard-version: [""]
```
tatiana, line 83: It's probably worth adding a comment on what this means here. I was confused when I first read this part of the code.

dagfactory/dagbuilder.py
```diff
@@ -47,12 +47,51 @@
...
+    )
+    from airflow.sensors.python import PythonSensor as _CorePythonSensor
+
+    _python_operator_classes.append(_CorePythonOperator)
+    _branch_python_operator_classes.append(_CoreBranchPythonOperator)
+    _python_sensor_classes.append(_CorePythonSensor)
+except ImportError:  # pragma: no cover
+    pass
+
+if not _python_operator_classes:  # pragma: no cover
```
tatiana, line 90: I believe we could make this message more targeted. My understanding is that Airflow 2 has `airflow.operators.python`, so this scenario should not happen in Airflow 3. Or... did Airflow 2 introduce a breaking change and remove these built-in operators?

I feel like the message should be more targeted and explicitly tell that Airflow 3 expects users to install `apache-airflow-providers-standard ` to be able to use PythonOperator/BranchPythonOperator/PythonSensor

dagfactory/dagbuilder.py
```diff
@@ -255,7 +294,7 @@ def make_task(operator: str, task_params: Dict[str, Any]) -> BaseOperator:
         # class is a Callable https://stackoverflow.com/a/34578836/3679900
         operator_obj: Callable[..., BaseOperator] = import_string(operator)
         # pylint: disable=too-many-nested-blocks
-        if issubclass(operator_obj, (PythonOperator, BranchPythonOperator, PythonSensor)):
+        if issubclass(operator_obj, PYTHON_CALLABLE_CLASSES):
```
pankajastro, line 343: The import changes at top is looking a bit complex I was thinking if the subclass is the only issue can we do something like

Probably at top of this file

```python
_PYTHON_CALLABLE_PATHS = frozenset({                                                                                                                     
      "airflow.operators.python.PythonOperator",
      "airflow.operators.python.BranchPythonOperator",                                                                                                     
      "airflow.sensors.python.PythonSensor",                                                                                                               
      "airflow.providers.standard.operators.python.PythonOperator",
      "airflow.providers.standard.operators.python.BranchPythonOperator",                                                                                  
      "airflow.providers.standard.sensors.python.PythonSensor",
  })
```

then here check 
```python
if operator in _PYTHON_CALLABLE_PATHS:
```

tatiana (approved):
Thanks for fixing this, @pankajkoti . It feels weird that Airflow 2 users can install `apache-airflow-providers-standard`, but since this affects a customer, we need to address this use case.

dagfactory/dagbuilder.py
pankajkoti: Thanks @pankajastro. I considered the path-set approach but kept the class union so `issubclass` still triggers `python_callable` resolution for user-defined subclasses of `PythonOperator` (e.g., an internal wrapper). A fixed frozenset of dotted paths would silently skip those.

Expanded the comment at the import block in [6ee1331](https://github.com/astronomer/dag-factory/pull/736/commits/6ee1331b91332e721e6b8a635d7bca8920d7d609) to call out this rationale.

dagfactory/dagbuilder.py
pankajkoti: Elaborated on the message and also added comments above for this approach

.github/workflows/cicd.yaml
pankajkoti: yes added a comment.

PR #739, opened 2026-05-05
Fix sys.path for DAG helper modules in Astro bundle environments

pankajastro:
Follow-up: https://github.com/astronomer/dag-factory/pull/613

In Astro Hosted, DAG bundles deploy to a timestamped path (/tmp/airflow/dag_bundles/.../dags/) that is not in sys.path, so helper modules (hacker_news, pypi_stats, sample) are not importable. Each file now inserts its own directory via __file__ so the path resolves correctly regardless of the deployment timestamp.

### DIFF
--- a/dev/dags/comparison/example_hackernews_plain_airflow.py
+++ b/dev/dags/comparison/example_hackernews_plain_airflow.py
@@ -1,8 +1,10 @@
 import sys
 from pathlib import Path
 
-# Astro DAG bundles deploy to a timestamped path not in sys.path; insert parent dags dir so helper modules are importable.
-sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+# Astro DAG bundles deploy to a timestamped path not in sys.path; add parent dags dir so helper modules are importable.
+dags_parent = str(Path(__file__).resolve().parent.parent)
+if dags_parent not in sys.path:
+    sys.path.append(dags_parent)
 
 from datetime import datetime
 
--- a/dev/dags/comparison/example_hackernews_plain_airflow.py
+++ b/dev/dags/comparison/example_hackernews_plain_airflow.py
@@ -1,3 +1,9 @@
+import sys
+from pathlib import Path
+
+# Astro DAG bundles deploy to a timestamped path not in sys.path; insert parent dags dir so helper modules are importable.
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
 from datetime import datetime
 
 try:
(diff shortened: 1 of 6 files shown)


dev/dags/comparison/example_hackernews_plain_airflow.py
```diff
@@ -1,3 +1,11 @@
+import sys
+from pathlib import Path
+
+# Astro DAG bundles deploy to a timestamped path not in sys.path; add parent dags dir so helper modules are importable.
+dags_parent = str(Path(__file__).resolve().parent.parent)
```
pankajkoti: we call it dags_parent here and in other files config root dir. Can we have a consistent name across the dags?

dev/dags/comparison/example_hackernews_plain_airflow.py
```diff
@@ -1,3 +1,11 @@
...
+if dags_parent not in sys.path:
+    sys.path.append(dags_parent)
```
pankajkoti: Why do we use `append` here and not `insert(0,..)` like in other dags. Can we make it consistent?

dev/dags/comparison/example_pypi_stats_plain_airflow.py
```diff
@@ -1,5 +1,13 @@
 from __future__ import annotations
 
+import sys
+from pathlib import Path
+
+# Astro DAG bundles deploy to a timestamped path not in sys.path; add parent dags dir so helper modules are importable.
+dags_parent_dir = str(Path(__file__).resolve().parent.parent)
+if dags_parent_dir not in sys.path:
+    sys.path.append(dags_parent_dir)
```
pankajkoti: same questions about using config root dir and insert(0,..) instead of append

pankajkoti (approved):
LGTM. A couple of suggetions around consistency

dev/dags/comparison/example_hackernews_plain_airflow.py
pankajastro: changed to `config_root_dir_str`

dev/dags/comparison/example_hackernews_plain_airflow.py
pankajastro: yes, changed to `insert(0,..)`

dev/dags/comparison/example_pypi_stats_plain_airflow.py
pankajastro: Fixed

PR #740, opened 2026-05-07
docs: align contributor and quick-start docs with current Python/Airflow support

pankajastro:
The howto referenced an Airflow 2.5 hatch env that no longer exists and capped Python at 3.12 over a stale `google-re2` 3.13 wheel issue; the Astro quick-start still listed Python 3.9. Update them to match the current supported range (Python 3.10–3.14, Airflow 2.9+).


tatiana (approved):
Thanks for updating the docs, @pankajastro , fixing these version inconsistencies!

PR #741, opened 2026-05-08
feat: add strict_mode flag for per-DAG error isolation

corsettigyg:
Re-opened https://github.com/astronomer/dag-factory/pull/719 after messing up with the git history and addressing feedbacks

## Summary

- Introduces `AIRFLOW__DAG_FACTORY__STRICT_MODE` (default: `False`) that, when enabled, raises a `DagFactoryConfigException` for any DAG that fails to build — while still registering all successfully built DAGs.
- `build_dags()` now catches each DAG independently and returns `(dags, build_errors)`, so a single broken DAG in a multi-DAG YAML never prevents the others from loading.
- In folder-scan mode, each YAML file is processed independently — a broken file does not prevent other files from loading.

## Motivation

DAGs silently disappearing from the Airflow UI when a YAML config has a parsing error is hard to debug. Strict mode surfaces these errors as Airflow "Broken DAG" import errors visible in the UI and CI/CD pipelines, while still loading all valid DAGs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,24 +6,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
-
-### Added
-
-- Add `AIRFLOW__DAG_FACTORY__STRICT_MODE` (`dag_factory.strict_mode`, default `False`).
-  When enabled, raises `DagFactoryConfigException` for any DAG that fails to build, while still
-  registering all DAGs that built successfully. In folder-scan mode, a broken YAML file no longer
-  prevents other files from loading.
-
-### Internal
-
-- `_DagFactory.build_dags()` now returns `Tuple[Dict[str, DAG], Optional[Tuple[str, Exception]]]`
-  instead of `Dict[str, DAG]`. `_DagFactory` is a private class (underscore-prefixed, not exported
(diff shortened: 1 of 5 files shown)


corsettigyg:
hey @pankajastro , I ended up closing the previous PR and re-opening here since I messed up the git history a little bit haha addresing your previous points

> @corsettigyg, thanks for working on this. Do we need to make changes to the build_dags() and _generate_dags() functions? Since we encourage users to use the load_yaml_dags function for DAG authoring, couldn’t the strict check logic live only in that function? Also, we’re planning to release a new version in a couple of days, so if we can wrap up the review feedback, we can include it in the upcoming release.

since the errors are captured during the dag contruction, I cant think of a reliable way of doing it without touching the build_dags or _generate_dags. we need the DAG loop to capture it

> I would call it a breaking change (the return type change). We should document it in the changelog.

`_DagFactory` is a private method so it should not be required tbh, but open to document it if needed

> Because of the try/except block, if a user was directly using the method, it will no longer raise the error and will instead return it. We may need to document this or reduce its scope of how we are propagating strict mode error.

Falls back to the private method argument. I'd expect no one to be using private methods in the code

> why are you appending error here without checking strict_errors?
> why are you storing all error and raising only the first error's traceback

This was part of an old attempt i tried doing of raising all errors in the airflow UI, but it didnt work out since there is a char limit in the UI. I get the first error traceback to visually break things in the UI without overloading the char limit, so the user knows something is wrong. I wanted to do it for every single YAML but implementation was much more complex. open for suggestions tho

pankajastro (approved):
LGTM

PR #849, opened 2026-07-24
fix: validate Python callable file and name together

### LINKED ISSUE #101: Looks like a bug in dagbuilder.py
The condition on 197 looks like a bug. As both are required, there should be **or** instead of **and**. 
image

cen-zp:
## Summary

- validate that `python_callable_name` and `python_callable_file` are provided together
- raise a clear `DagFactoryException` for either incomplete configuration
- cover both missing-pair cases and tighten the existing exception assertion

## Root cause

The builder only rejected configurations where all Python callable options were absent. When exactly one file-based callable parameter was provided, it accessed the missing dictionary key directly and raised `KeyError`.

## Validation

- `hatch run tests.py3.11-3.2:test` — 295 passed, 7 skipped, 1 deselected
- `pytest tests/test_dagbuilder.py -q` — 87 passed, 4 skipped
- applicable pre-commit checks on the changed files

Fixes #101

### DIFF
--- a/dagfactory/dagbuilder.py
+++ b/dagfactory/dagbuilder.py
@@ -354,6 +354,13 @@ class DagBuilder:
                     "from a file. with the special pyyaml notation:\n"
                     "  python_callable_file: !!python/name:my_module.my_func"
                 )
+            if not task_params.get("python_callable") and bool(task_params.get("python_callable_name")) != bool(
+                task_params.get("python_callable_file")
+            ):
+                raise DagFactoryException(
+                    "Failed to create task. `python_callable_name` and "
+                    "`python_callable_file` must be provided together."
+                )
             if not task_params.get("python_callable"):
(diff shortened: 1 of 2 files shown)


tatiana (commented):
@cen-zp Thanks for the fix, and nice call not just flipping the `and` to `or` as the issue suggested - that would have broken the valid "python_callable only" and "name + file" configs.  Doing an XOR check to the `python_callable`-absent case feels like the right approach,

One optional cleanup: the new check repeats `not task_params.get("python_callable")` right before the existing `if not ...:`, so it reads as three back-to-back `python_callable` checks. 

I'm happy to approve and merge it once you change it out of draft!

cen-zp:
Hi @tatiana, thanks again for the review! I’ve applied the optional cleanup and marked the PR ready for review. The latest workflow runs also appear to be awaiting maintainer approval. Could you please take another look when you have a chance?

PR #869, opened 2026-08-18
feat: add dag_codegen module and generate CLI command

### LINKED ISSUE #868: [Feature] Add a `generate` CLI command to pre-compile YAML DAGs into static Python files
### Description

`dag-factory` currently builds DAGs by parsing YAML at runtime, every time the DAG processor scans the `dags_folder` (`load_yaml_dags`). This works well, but at scale (many DAGs, large YAML files) it adds real, repeated overhead to DAG-parsing time — the DAG processor re-interprets and rebuilds every DAG on every scan cycle, whether anything changed or not.

This proposes adding a `dagfactory generate <yaml_dir> <py_dags_dir>` CLI command that pre-compiles YAML DAG configs into static, plain Airflow `.py` files — once, at build/CI time — instead of interpreting YAML on every DAG-processor cycle. The generated files are ordinary Python: no runtime dependency on dag-factory's YAML loader, just `DAG(...)` / operator constructor calls that Airflow parses like any hand-written DAG.

### What it supports

Verified by generating and **executing** the output against dag-factory's own example DAGs (21 of 24 real example YAMLs generate and run correctly out of the box — the remaining 3 fail identically with the existing runtime loader too, for unrelated reasons: a missing optional provider package, and pre-existing bugs in the example YAMLs themselves):

- Regular operators, `dependencies` (`>>`), `default_args` (inline and via a shared external `defaults.yml`)
- `task_groups` (nested, list or mapping YAML syntax, group-level `dependencies` and `default_args`/callbacks)
- TaskFlow (`decorator:`) tasks, including the `+task_name` upstream-output reference convention, with real signature introspection to route kwargs between the `@task` decorator and the wrapped callable (matching `make_decorator`'s `inspect.getfullargspec` behavior)
- Dynamic task mapping (`expand`/`partial`) for both regular operators and TaskFlow tasks, including `.output`/`XcomArg(...)` and `+task_name` mapped-value references
- `python_callable_name`/`_file`, and the equivalent for HTTP sensor `response_check_*` / SqlSensor `success_check_*`/`failure_check_*`
- Callbacks via `<type>_name`/`<type>_file`, plain dotted-string callables, and the dict form (`{callback: "...", **params}` → `functools.partial(...)` or a direct Notifier instantiation, matching `set_callback` exactly — not left as an inert dict)
- Custom Python objects (`__type__`/`__args__` YAML tags, e.g. `Asset`, custom `Timetable`s)
- Dataset/Asset schedules, including boolean conditions (`&`/`|`) and file-based dataset maps, delegated to `DagBuilder`'s own resolution logic rather than reimplemented
- `variables_as_arguments` (via `Variable.get(...)`, evaluated at DAG-parse time — deliberately not baked in at generation time, since Variable values live in the metadata DB)
- `doc_md_file_path` / `doc_md_python_callable_name`+`_file`

**Not supported**, and explicitly rejected with a clear error (falls back to the runtime loader) rather than silently producing broken output:
- `schedule_interval` (Airflow 2 syntax) — same rejection message as the runtime loader already gives under Airflow 3
- Task-level combinations the generator can't safely resolve without live introspection

The CLI fails the build (non-zero exit) if any DAG or file is skipped for any reason, specifically so a CI pipeline can't silently ship fewer DAGs than intended.

### Use case/motivation

This is the building block for a documented CI/CD pattern: run `dagfactory lint` + `dagfactory generate` in CI before deploying, ship only the generated `.py` files to the `dags_folder`, and let the DAG processor parse plain Python instead of re-interpreting YAML on every cycle. Today there's no supported way to do this, so teams hitting DAG-parsing overhead at scale are left writing ad hoc pre-generation scripts instead of a documented, tested approach.

### Related issues

_No response_

- [x] Yes — I have a complete, tested implementation ready (unit tests + verified against dag-factory's own example DAGs by executing the generated output, not just asserting on generated text). Will open the PR linked to this issue.

matiasmorenoiglesias:
# Title

feat: add `generate` CLI command to pre-compile YAML DAGs into static Python files

# Descripción (cuerpo del PR)

Closes #868

## Summary

Adds a `dagfactory generate <yaml_dir> <py_dags_dir>` CLI command that pre-compiles YAML DAG configs into static, plain Airflow `.py` files at build time, instead of relying on `load_yaml_dags()` to re-parse YAML on every DAG-processor scan cycle. The generated files have no runtime dependency on dag-factory's YAML loader — they're ordinary `DAG(...)`/operator constructor calls that Airflow parses like any hand-written DAG.

This is the building block for a CI/CD pattern: run `dagfactory lint` + `dagfactory generate` before deploying, ship only the generated `.py` files to the `dags_folder`, and let the DAG processor parse plain Python instead of re-interpreting YAML on every cycle.

## What's supported

Verified by generating **and executing** the output against dag-factory's own example DAGs — 21 of 24 real example YAMLs in `dev/dags/` generate and run correctly out of the box. The remaining 3 fail identically with the existing runtime loader too (confirmed by running both loaders against the same files), for reasons unrelated to this PR: a missing optional provider package, and two pre-existing bugs in the example YAMLs themselves.

- Regular operators, `dependencies` (`>>`), `default_args` (inline and via a shared external `defaults.yml`)
- `task_groups` — nested, both list and mapping YAML syntax, group-level `dependencies` and `default_args`/callbacks. Nesting is resolved by nesting generated `with TaskGroup(...):` blocks (Airflow resolves `parent_group` from context automatically, no explicit kwarg needed)
- TaskFlow (`decorator:`) tasks, including the `+task_name` upstream-output reference convention. Kwargs are routed between the `@task` decorator and the wrapped callable using real signature introspection (`inspect.getfullargspec`) at generation time, matching `make_decorator`'s own routing
- Dynamic task mapping (`expand`/`partial`) for both regular operators and TaskFlow tasks, including `.output`/`XcomArg(...)` and `+task_name` mapped-value references
- `python_callable_name`/`_file`, and the equivalent for HTTP sensor `response_check_*` / SqlSensor `success_check_*`/`failure_check_*`
- Callbacks via `<type>_name`/`<type>_file`, plain dotted-string callables, and the dict form (`{callback: "...", **params}`) — rendered as `functools.partial(...)` or a direct Notifier instantiation, matching `set_callback` exactly, never left as an inert dict
- Custom Python objects (`__type__`/`__args__` YAML tags — e.g. `Asset`, custom `Timetable`s)
- Dataset/Asset schedules, including boolean conditions (`&`/`|`) and file-based dataset maps — delegated to `DagBuilder`'s own resolution logic rather than reimplemented
- `variables_as_arguments`, rendered as `Variable.get(...)` evaluated at DAG-parse time (deliberately not baked in at generation time, since Variable values live in the metadata DB, not an importable module)
- `doc_md_file_path` / `doc_md_python_callable_name`+`_file` — set on the DAG object after construction, matching dagbuilder (these aren't `DAG()` constructor kwargs)

**Not supported**, and explicitly rejected with a clear error rather than silently producing broken output:
- `schedule_interval` (Airflow 2 syntax) — same rejection message the runtime loader already gives under Airflow 3
- Any task-level combination the generator can't safely resolve without deeper live introspection than is currently implemented

The CLI fails the build (non-zero exit code) if any DAG or file is skipped for any reason — a CI pipeline running `generate` can't silently ship fewer DAGs than intended.

## Design notes

- `task_groups` and dynamic task mapping are implemented as a small composite/recursive renderer (`render_task_group_block`), not special-cased per depth — arbitrary nesting works because each block is generated as self-relative source and uniformly re-indented by the caller (`indent_block`), rather than hardcoding indentation per nesting level.
- For anything genuinely complex to re-implement safely as static source (dataset boolean-condition parsing, callback resolution from name+file), the generated code calls dag-factory's own existing runtime helpers (`DagBuilder.evaluate_condition_with_datasets`, `get_python_callable`, etc.) rather than duplicating that logic — single source of truth, less risk of drift.
- All string values are rendered via `repr()`, not hand-built quoting, so values containing quotes/newlines/backslashes don't produce invalid generated source.
- Tasks referencing another task's output inline (TaskFlow `+task_name`, `expand`'s `.output`) are topologically ordered in the generated source, independent of their order in the YAML — dag-factory doesn't require YAML tasks to be topologically ordered, and the generated Python has to preserve that.

## Test plan

- [x] Unit tests for every helper (`build_task_group_tree`, `resolve_taskflow_value`, `resolve_expand_value`, `get_callable_arg_names`, `resolve_callback_dict`, etc.) — see `tests/test_dag_codegen.py` and additions to `tests/test__main__.py`
- [x] Generated output executed (not just asserted on as text) against all applicable example YAMLs in `dev/dags/`
- [x] Cross-checked against the existing runtime loader (`load_yaml_dags`) for parity on `task_groups`, defaults merging, and the 3 example YAMLs that fail — same failures, same root cause, on both paths
- [x] Full existing test suite passes unchanged (excluding one pre-existing failure from a provider package not installed in the dev environment, unrelated to this change)


matiasmorenoiglesias:
Rebased this branch on top of the latest `main` (140 commits had landed since I branched off) to make sure this merges cleanly with no conflicts.

Two things came up during the rebase that are worth flagging explicitly, since they're environment-driven rather than logic bugs in the original implementation:

1. **`airflow.utils.module_loading` no longer exists** on the Airflow version now pinned by `main` (3.3.1) — `dag_codegen.py` was importing `import_string` from there directly. Fixed by reusing the same `import_string` that `dagbuilder.py` already exposes (it has its own version-compatible fallback for this exact import path), instead of duplicating that fallback logic.

2. **A new example DAG landed on `main`** since I branched off — `dev/dags/example_multiple_outputs_taskflow.yml`, exercising `multiple_outputs=True` with the `+task["xcom_key"]` reference syntax (e.g. `value: "+collect['key1']"`). `dag_codegen` wasn't handling that bracket/subscript form yet — it was mangling it into an invalid identifier (`collect['key1']` → `collect__key1__`). Fixed by matching the exact same regex `dagbuilder.py`'s `_replace_kwargs_values_as_xcom` uses (`^\+(?P<task>\w*)(\[['\"](?P<xcom_key>.*)['\"]\])?$`) when resolving TaskFlow `+`-prefixed values, so `+collect['key1']` now correctly generates `collect['key1']` (a real subscript on the task's XComArg-returning variable) instead of garbage.

Re-validated after the rebase:
- Full existing test suite passes (one pre-existing failure unrelated to this PR: a provider package not installed in the dev environment).
- Generated output for all 26 example YAMLs under `dev/dags/` (including the 2 new ones added upstream) executed directly — 22 run correctly, and the 4 that don't fail identically with the existing runtime YAML loader too (confirmed by running both loaders side by side), for reasons unrelated to this PR: a missing optional provider package, and pre-existing issues in the example YAMLs themselves.

Branch is up to date with `main` and ready for review.

@pankajastro @pankajkoti

matiasmorenoiglesias:
@pankajkoti  
Is there anything you can do or help with?

matiasmorenoiglesias:
@pankajastro
Quedo atento a algun comentario, saludos

